PR #1, opened 2016-08-12
Change delete button style to make it more visible

gobindaakhuli:
On card-list page delete button is working but it had not any effect for active this button.
Means when mouse pointer go through on this button it was not indicate that it was active.

So in style sheet I did added hover effect for this delete button, and now it show it active when mouse pointer go through on this button.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/1)

### DIFF
--- a/sample.css
+++ b/sample.css
@@ -4,3 +4,8 @@ button.delete-button{
   background-color: gray !important;
   color:white;
 }
+button.delete-button:hover{
+  color: black;
+  font-style: italic;
+  font-weight: bold;
+}


kwk:
@gobindaakhuli thank you for this contribution but it is probably not wise to fiddle with CSS right now as we try to use a framework / library for this.

joshuawilson:
@gobindaakhuli Thanks for the PR. However this project was just created and the UI layout that you changed is changing to be built using PatternFly. If you do want to contribute to this project, that would great. Either create an issue or comment on an existing one and if no one has started it, you can work on that.

joshuawilson:
@gobindaakhuli if you take a look now, we have some of the designs in place. I'm looking forward to some PRs from you.

PR #12, opened 2016-08-22
Fix REST path to workitems

kwk:
In https://github.com/almighty/almighty-core/issues/97 we've changed the core to list workitems under `/api/workitems` (notice the "s" at the end). The docker container of almighty-core is not using this at the moment but we try to get it in there as quick as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/12)


sbose78:
Looks good.

PR #13, opened 2016-08-23
Remove use of Red Hat prefix

maxandersen:
Why:

 * Red Hat ALM is not a name registered or used.

This change addreses the need by:

 * Put "Almighty" instead of "Red Hat ALM" for now.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/13)


joshuawilson:
Thanks for the PR. Merged!

Issue #8, opened 2016-08-17
As a user I should be able to login into the application.


joshuawilson:
The login page should be a component version of the PatternFly style. We should be able to use the HTML and CSS in the PatternFly example.

tsmaeder:
With Nimisha...

joshuawilson:
@aslakknutsen can you please update this issue when the backend supports the Login?

michaelkleinhenz:
I think this is a story and should be split up into tasks. I can't see what's the progress here.

nimishamukherjee:
PR #189 merged with master, hence closing this story.

PR #14, opened 2016-08-24
Added login service

nimishamukherjee:
Added a login service - #8

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/14)


joshuawilson:
I had to do a bit of merging to fit it in with the latest but it looks like it is fine.

PR #17, opened 2016-08-26
Added initial POC test for UI - test implemented in SelenIDE and Java…

ldimaggi:
… to serve as POC for UI tests

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/17)


bartoszmajsak:
I'm not sure if we have to build our own framework for that (that's what I see here in this PR as a foundation for it). We have [Arquillian Drone](https://github.com/arquillian/arquillian-extension-drone) and [Graphene](https://github.com/arquillian/arquillian-graphene) (linked in the README btw), which we can build upon (and extend if needed). I think @matousjobanek would be thrilled to see this happening. There is a lot of cool stuff we built over last few weeks with @lordofthejars - for example transparent usage of dockerized browser. It would be really great to have this project as a way to improve our tooling even further.

That said I will try to invest some time next week to contribute to this PoC and see how we can use our own tools for testing.

As for Selenide - it's really nice DSL for interacting with the elements and asserting on them. I would definitely like to have such a thing when writing UI tests. So big +1 from my side.

lordofthejars:
I think that if you are using Graphene then it has no sense to use
SelenIDE. If not using Graphene then it is worth to have it

MatousJobanek:
Hi,
yeah, I do agree with Alex. Arquillian Drone helps you more likely with 
-the lifecycle of a browser, 
-easy switching between types of browsers
-running more instances at the same time. 
It is also much easier to integrate it with Arquillian deployments and with Graphene and it offers many other things.
Looking at the Readme (as I don't know selenide) this tool is mainly focused on validation and checking correctness of displayed pages - which is basically the same as Arquillian Graphene does. So, Arquillian Graphene and Selenide seem to be two very similar tools for me. 
There are maybe some functions implemented in better way than it is in Graphene, but it's also true in the opposite direction. If there is something that you are missing there in Graphene or Drone, then I'm completely open to a discussion about implementing new features or modifying the current ones.
From my point of view, the biggest benefits of Graphene and Drone are:
- they are Red Hat projects whose implementation and releases we can control based on our needs
- it has also wide usage in RH QE teams
- it offers big opportunity for integrating with other Arq modules, etc...

It's quite surprising that you're mentioning Selenide as your main tool for UI testing, but at the same time in the same document you're referencing Graphene pages with abstraction that are already implemented there in Graphene project. 
May I know why you have decided for Selenide?

bartoszmajsak:
> I think that if you are using Graphene then it has no sense to use
SelenIDE. If not using Graphene then it is worth to have it

> So, Arquillian Graphene and Selenide seem to be two very similar tools for me. 

That is not entirely true, as SelenIDE is a mix of what Graphene and Fleuntlenium is. It's not only a simplification for querying / interacting with web elements (as Graphene mainly is I guess), but has also handy assertions, which, correct me if I'm wrong, we do not have in place in Graphene, do we?

bartoszmajsak:
Let's have a chat all together on Monday to discuss all those hot topics and how we can all work together to make it rock solid. I will set up something @ldimaggi @lordofthejars @MatousJobanek @joshuawilson

lordofthejars:
Bartosz you are right about assertions, but we can integrate them in
Graphene

bartoszmajsak:
Sure we can, but I'm not sure if we should :)

PR #18, opened 2016-08-26
SASS-Loader added. Board component + footer component changed to SASS

SMahil:
SASS files added to board component and footer component.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/18)


joshuawilson:
@nimishamukherjee Can you please review this too. I think we all need to be on the same page before we bring this into the project.

joshuawilson:
Should we leave the CSS loader and support in place too? We could have the loader support CSS, LESS, and SASS. I don't know if that would make things more confusing but it is possible.

nimishamukherjee:
@joshuawilson , @SMahil - Patternfly scss files are not available under the patternfly node package. That only has the css files. For now we can go ahead with the approach of including the patternfly files under the assets folder.

joshuawilson:
@nimishamukherjee I think that will work for now. Thanks for reviewing this.

mindreeper2420:
PatternFly SASS is available via the GEM package manager 

https://rubygems.org/gems/patternfly-sass

joshuawilson:
@mindreeper2420 do you know how to make that work with Webpack? We are trying to keep our dependencies handled by NPM. I'll have to consider using the gem.

joshuawilson:
Reviewed 18 of 18 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/almighty/almighty-ui/18)*

PR #22, opened 2016-09-01
Variable type matching issue fixed

sanbornsen:
Why?
 - The update page was not getting any data from the service
 - In the service after fetching the data it was finding against a number type variable in string type variables

What?
- In the find function we had to do a compare instead of strict compare

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/22)

### DIFF
--- a/src/app/work-item/work-item.service.ts
+++ b/src/app/work-item/work-item.service.ts
@@ -22,7 +22,7 @@ export class WorkItemService {
       .catch(this.handleError);
   }
 
-  getWorkItem(id: number): Promise<WorkItem> {
+  getWorkItem(id: string): Promise<WorkItem> {
     return this.getWorkItems()
       .then(workItems => workItems.find(workItem => workItem.id === id));
   }
--- a/src/app/work-item/work-item-detail/work-item-detail.component.html
+++ b/src/app/work-item/work-item-detail/work-item-detail.component.html
@@ -1,3 +1,4 @@
+<app-header></app-header>
 <div *ngIf="workItem">
   <h2>{{workItem.name}} details!</h2>
   <div>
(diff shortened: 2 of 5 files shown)


sbose78:
This looks good but you should not be opening a new PR.
Commiting to the same branch updates the existing PR :)

sanbornsen:
No, this is another change. I have 2 PR activated now for two different issues. 21 and 22

sbose78:
Cool, then! :)

src/app/work-item/work-item.service.ts
```diff
@@ -24,7 +24,12 @@ export class WorkItemService {
 
   getWorkItem(id: number): Promise<WorkItem> {
     return this.getWorkItems()
-      .then(workItems => workItems.find(workItem => workItem.id === id));
+      .then((workItems) => {
+ 		  // workItem.id is string 
+		  // but id is number 
+		  // so we use == instead ===
+		  return workItems.find(workItem => workItem.id == id);
+	  });
```
joshuawilson: Good find. 
In an arrow function, if you don't do anything else in the braces you do not need the 'return'. It is easier to read, when using arrow functions, on one line. I would not make it multiple lines with a return just to add comments. If you think the comments are needed then add them above the '.then'. I do think the comment is needed since the switch from === to == needs be explained.

joshuawilson:
Thanks for the fix. Can you please squash the 2 commits into one that says what you want it to say.

sanbornsen:
@joshuawilson does this work?

joshuawilson:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/almighty/almighty-ui/22)*

joshuawilson:
merged

PR #23, opened 2016-09-01
Quick add a Work Item Type(almighty/almighty-core#99 UI Issue 3).

nimishamukherjee:
When viewing a list of work items, the top row will be used to quickly add a work item.
When the item name is entered the add button (+) gets enabled. After adding, the list of work items are displayed in reversed order so that the item that was added is displayed at the top.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/23)

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.css
+++ b/src/app/work-item/work-item-list/work-item-list.component.css
@@ -16,3 +16,6 @@
 	padding: 1% 8%;
 }
 .error {color:red;}
+.supress-hover:hover{
+  background-color: transparent;
+}
--- a/src/app/work-item/work-item-list/work-item-list.component.html
+++ b/src/app/work-item/work-item-list/work-item-list.component.html
@@ -1,13 +1,10 @@
 <app-header></app-header>
-<div class="container-fluid container-cards-pf">
-<button class="btn btn-primary" (click)="addWorkItem()">Add New Work Item</button>
-<div *ngIf="addingWorkItem">
-  <work-item-quick-add (close)="close($event)"></work-item-quick-add>
-</div>
-<br /><br />
 <div class="container-fluid">
 	<div class="list-group list-view-pf list-view-pf-view">
-		<div *ngFor="let workItem of workItems">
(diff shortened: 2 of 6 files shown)


mindreeper2420:
@nimishamukherjee @joshuawilson FYI, this does not match the design specs from the UXD Team. It will need to be altered to match the proper specs. @mgranfie

nimishamukherjee:
@mindreeper2420 we have used the patternfly form component here. This is a temporary UI. Once the final quick add flow and screen designs are available, we will update this screen,

src/app/login/login.service.ts
```diff
@@ -8,7 +8,7 @@ import { LoginItem } from './login-item';
 @Injectable()
 export class LoginService {
   // private githubUrl = 'app/loginStatus';
-  private githubUrl = 'http://localhost:8081/api/login/authorize ';  // URL to web api
+  private githubUrl = 'http://localhost:8080/api/login/authorize ';  // URL to web api
```
joshuawilson: Why did you change this? It is set to 8081 so that it doesn't conflict with the backend when running on localhost.

joshuawilson:
What is the second commit for? It appears to deal with the Login and the Details pages.

joshuawilson:
@nimishamukherjee at the moment we need something that is better then we had. This commit fills in the gap. We know it is a throw away design.

src/app/work-item/work-item-detail/work-item-detail.component.html
```diff
@@ -1,31 +1,61 @@
...
+  <br />
+  <div class="form-horizontal">
+    <div class="form-group">
+      <label class="col-sm-1 control-label">ID</label>
+      <div class="col-sm-5">
+        {{workItem.id}}
+      </div>
+    </div>
+    <div class="form-group">
+      <label class="col-sm-1 control-label" for="name">Name</label>
```
joshuawilson: The use of 'for' does not have anything to refer to. You need to add an 'id' or 'name' attribute to the div the label is 'for'.

nimishamukherjee:
@joshuawilson removed the second commit.

PR #31, opened 2016-09-02
Set environment variables in dev and prod config files

nimishamukherjee:
Request from @kwk - "Can you configure the api endpoint of the core to be read from environment variables? Just as a hack so that we can have it in master and use that branch instead of some other branch?"
Based on https://github.com/AngularClass/angular2-webpack-starter/wiki/How-to-pass-environment-variables%3F - Updated the webpack.dev.js and webpack.prod.js files.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/31)

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -1,7 +1,16 @@
+var webpack = require('webpack');
 var webpackMerge = require('webpack-merge');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var commonConfig = require('./webpack.common.js');
 var helpers = require('./helpers');
+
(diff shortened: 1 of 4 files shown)


kwk:
Thanks @nimishamukherjee ! Your changes do work and I've rebased my demo branch onto your PR's branch. I don't know though how to decide whether development or production is run. Can you give me instructions on how this can be done? Is there some npm CLI switch or something?

nimishamukherjee:
@kwk I need @joshuawilson's input on this. I think we should change webpack.config.js to
module.exports = require('./config/webpack.prod.js');
I haven't tried the production mode.

joshuawilson:
merged

PR #32, opened 2016-09-02
Stuff that needs to be in master for the demo on sept 6th

kwk:
I will add a description later

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/32)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,4 +9,7 @@ COPY package.json /usr/src/app/
 RUN npm install
 COPY . /usr/src/app
 
-CMD [ "npm", "run", "docker:run" ]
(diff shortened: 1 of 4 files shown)


src/app/work-item/work-item.service.ts
```diff
@@ -8,9 +8,7 @@ import { WorkItem } from './work-item';
 @Injectable()
 export class WorkItemService {
   private headers = new Headers({'Content-Type': 'application/json'});
-  // private workItemUrl = 'app/workItems';  // URL to web api
```
joshuawilson: please don't remove this line as we need it for development

joshuawilson:
@kwk can you please rebase this onto the lastest master, the first commit is from the last PR.
Then can you please squash all the commits that are dealing with the same issue. Mostly I'm seeing changes to the `docker-compose` file.

PR #41, opened 2016-09-03
Fix Prod build

baijum:
`webpack.prod.js` seems to be not working. So, use `webpack.dev.js` temporarily.

The `-p` flag also not working, probably it should be added back later.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/41)


src/app/shared-component/status-drawer.component.ts
```diff
@@ -10,7 +10,7 @@ import { WorkItem } from './../work-item/work-item';
 @Component({
 	selector   : 'status-drawer',
 	templateUrl: './status-drawer.component.html',
-	styleUrls  : ['src/app/shared-component/status-drawer.component.css'],
+	styleUrls  : ['/status-drawer.component.css'],
```
sanbornsen: This seemed not to be working. Did you test this @baijum ? Also, why will not this path work on production? I believe it's a local reference to the app which is constant.

baijum: It's working for me. This change is consistent with rest of the code:

```
$ find . -name "*.ts"|xargs grep "css"
./app.component.ts:  styleUrls: ['/app.component.scss']
./footer/footer.component.ts:    styleUrls: ['/footer.component.scss'],
./login/login.component.ts:    styleUrls: ['/login.component.css'],
./board/board.component.ts:  styleUrls: ['/board.component.scss']
./app.module.ts:import '../../public/scss/styles.scss';
./shared-component/status-drawer.component.ts:  styleUrls  : ['/status-drawer.component.css'],
./work-item/work-item-list/work-item-list.component.ts:  styleUrls: ['/work-item-list.component.css'],
./work-item/work-item-detail/work-item-detail.component.ts:  styleUrls: ['/work-item-detail.component.css']
./work-item/work-item-search/work-item-search.component.ts:    styleUrls:  ['/work-item-search.component.css'],
./work-item/work-item-quick-add/work-item-quick-add.component.ts:  styleUrls: ['/work-item-quick-add.component.css']
```

sanbornsen: Hey but it was not working for me. I was getting an exception. Then I went through [this](https://github.com/angular/angular-cli/issues/1501) post. The issue was with the white space before ':'. Instead of styleUrls : [] I did styleUrls: [] and it did work. I wonder, how is it working for you though.

baijum: @sanbornsen I didn't check functionality, but there was no 404 error in the client :) Now I am away from my system and I can only check your pull request very late today.

baijum: @sanbornsen May be your are using old loader. In the new code, they add "." if it doesn't exist. See this https://github.com/TheLarkInn/angular2-template-loader/blob/master/index.js#L8 In the same file, line no. 3, you can see that they handle white-space also.

PR #57, opened 2016-09-04
WIP: Builder image

aslakknutsen:
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/57)


aslakknutsen:
test

Dockerfile.deploy
```diff
@@ -0,0 +1,6 @@
+FROM centos/nginx-16-centos7
```
baijum: Use tag to avoid any surprise. `FROM centos/nginx-16-centos7:1.6`

Dockerfile.builder
```diff
@@ -0,0 +1,25 @@
...
+
+# Setup the app directory
+ENV APP_DIR=/usr/src/app
+RUN mkdir -p ${APP_DIR}
+
+ENV PATH=${PATH}:/usr/local/node/node-v6.3.1-linux-x64/bin
+
+WORKDIR ${APP_DIR}
+
+RUN npm install bower
```
baijum: The command should be just `npm install`

PR #62, opened 2016-09-06
Adding scroll horizontally on work item list view

sanbornsen:
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/62)

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.css
+++ b/src/app/work-item/work-item-list/work-item-list.component.css
@@ -1,5 +1,5 @@
 .screen-min-ht{
-  min-height: 85%;
+  height: 80%;
 }
 .type span{
   font-size: 12px;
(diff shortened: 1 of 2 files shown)


joshuawilson:
Why do we want to force a scroll bar to appear even when we don't need it? 

Is there an issue # for this?

@mindreeper2420 Can you please review this?

mindreeper2420:
@joshuawilson @sanbornsen Having a scrollbar appear when it is not necessary gives the false impression of a scrollable interface, which can cause confusion and doubt. The browser should automatically handle any scrollbars (Chrome, IE, Firefox and Edge all handle them differently, including persistent view states).

If the List View is requiring a horizontal scrollbar (as stated here), then it is not sizing properly for the viewport. That should be done in lieu of forcing scrollbars.

Of note, the finalized design will makes this item non-applicable.

aslakknutsen:
Also not having it makes the page jump around moving from page to page. e.g. in demo.almighty.io view the List then go to board and back ... pick your poison I guess.

sanbornsen:
I am sorry. Excuse my limited knowledge in CSS. It should be good now.

PR #65, opened 2016-09-06
Added new environment value 'inmemory' that enables the in memory dat…

michaelkleinhenz:
Enables a new ENV value 'inmemory' that enables the in memory datasource. No files must be changed to enable inmemory anymore. Default is 'development' as before.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/65)

### DIFF
--- a/src/app/in-memory-data.service.ts
+++ b/src/app/in-memory-data.service.ts
@@ -1,7 +1,7 @@
 import { InMemoryDbService } from 'angular2-in-memory-web-api';
 export class InMemoryDataService implements InMemoryDbService {
   createDb() {
-    let workItems = [
+    let workitems = [
       {"fields":{"system.owner":"tmaeder","system.state":"open"},"id": "21", "name": 'Bullet Chart', description: 'Bullet Chart - Conceptual Design',"workItemType":"1","version":0 , status:'To Do', statusCode:0, type: 'story'},
       {"fields":{"system.owner":"tmaeder","system.state":"open"},id: 22, name: 'New Account', description: 'Create a New Account',"workItemType":"1","version":0 , status:'In Progress',statusCode:1, type: 'story'},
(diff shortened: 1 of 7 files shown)


src/app/in-memory-data.service.ts
```diff
@@ -1,7 +1,7 @@
 import { InMemoryDbService } from 'angular2-in-memory-web-api';
 export class InMemoryDataService implements InMemoryDbService {
   createDb() {
-    let workItems = [
+    let workitems = [
```
kwk: I'd leave this as `workItem`.

src/app/work-item/work-item.service.ts
```diff
@@ -8,18 +9,19 @@ import { WorkItem } from './work-item';
...
+    logger.log("WorkItemService using url " + this.workItemUrl);
+  }
 
   getWorkItems(): Promise<WorkItem[]> {
     return this.http
       .get(this.workItemUrl)
       .toPromise()
-      // .then(response => response.json().data as WorkItem[])
-      .then(response => response.json() as WorkItem[])
+      .then(response => process.env.ENV!='inmemory'?response.json() as WorkItem[]:response.json().data as WorkItem[])
```
kwk: This looks very packed but is not very readable.

kwk:
Except for some spelling things this LGTM.

src/app/in-memory-data.service.ts
michaelkleinhenz: I also would, but the name gets reflected as the url component and thus must be compatible with the "real" api.

src/app/work-item/work-item.service.ts
michaelkleinhenz: +1
But more readable than declaring an external func :-)

joshuawilson:
I'm not sure how long we will keep the in-memory data set, but this is a nice addition to use it.

PR #67, opened 2016-09-07
General API requests module created

sanbornsen:
What -
 * This is a single API request module built using angular2's http.
 * From now on we can use this module instead of direct http in services for API requests.

Why -
 * With the growing code base we'll have n number of services with time.
 * In future if we decide to use something else other than angular's http we just have to edit one file.
 * To remove code redundency.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/67)

### DIFF
--- a/src/utils/requests.ts
+++ b/src/utils/requests.ts
@@ -1,0 +1,58 @@
+import { Injectable } from "@angular/core";
+import { Headers, Http } from "@angular/http";
+
+@Injectable()
+export class Requests {
+
+	private headers = new Headers({'Content-Type': 'application/json'});
+	
+	constructor(private http: Http) { }
+
+	get(url: string, data: Object = {}): Promise<any> {
+		let ret:string[] = [];
+		for (var d in data){
+			ret.push(
+				encodeURIComponent(d) + "=" + encodeURIComponent(data[d])
+			);
+		}
+		let fullUrl: string = url;
+		if (ret.length)fullUrl += "?"+ret.join("&");
+		console.log(fullUrl);
+		
+		return this.http
+			.get(fullUrl)
+			.toPromise()
+			.then(response => process.env.ENV!='inmemory'?response.json() :response.json().data)
+			.catch(this.handleError);
+	}
+
+	delete(url: string): Promise<any> {
+		return this.http
+			.delete(url, {headers: this.headers, body: ""})
+			.toPromise()
+			.then(() => null)
+			.catch(this.handleError);
+	}
+
+	post(url: string, data: Object = {}): Promise<any> {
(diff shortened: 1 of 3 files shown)


sanbornsen:
@joshuawilson  @aslakknutsen please review.

src/utils/requests.ts
```diff
@@ -0,0 +1,58 @@
+import { Injectable } from "@angular/core";
+import { Headers, Http } from "@angular/http";
+
+@Injectable()
+export class Requests {
+
+	private headers = new Headers({'Content-Type': 'application/json'});
+	
+	constructor(private http: Http) { }
```
aslakknutsen: Maybe constructor should take a set of 'Additional Headers', so we can e.g. pass in the Authentication Token to be used.

sanbornsen: Good idea. Let me figure out how to do that.

sanbornsen: Angular2 simply takes over to the constructor. Looks like I have to find a way around.

sanbornsen: Looks like the only way to do that is using @Inject. Can we merge it for now, and later experiment more on it?

sanbornsen:
@aslakknutsen have a look at the second commit. This is not nicer, but this is the only way I could find to pass external parameters in a provider's constructor. It looks like Angular2 only allows injectors to instantiate another provider in a constructor. So I had to create a provider class Keys and using injector I instantiate it in Requests constructor. I tried almost all the possible way other than this, nothing seems work. **This commit is just for testing purpose**. Once you are done testing, I can reset the head for now.

CC: @joshuawilson

aslakknutsen:
@sanbornsen @nimishamukherjee is doing this as part of the Login work. Not sure what her solution is, but.. ?

sanbornsen:
Yes, so she is doing the same thing. She has created an authentication service provider, just like I created a keys provider for an example.

joshuawilson:
@michaelkleinhenz can you please review this PR?

michaelkleinhenz:
I think this should be done as a service. I am also always careful when introducing new layers of abstraction. In this case, it might be reasonable to have something like this, but I would definitely wait until we have a stable understanding on how authentication will work. As Aslak mentioned, this will definitely have an impact on how we should do things here.

sanbornsen:
We can make it as a service, and every other service will use it as a Request Service. Actually it is a service, just it doesn't follow the naming and positioning conventions. 
As I replied earlier in this thread, yes we can pass multiple parameters i.e. headers inside the constructor for authentication and any other operation using another service provider and instantiate it into this service's constructor.

PR #68, opened 2016-09-07
Listentry as component

michaelkleinhenz:
Extracted the list entry as a seperate component, making it reusable and self-contained (child widgets like buttons/menus are now controlled from the list entry itself, not from the parent). Wiring up, cleaning the wiring a bit. Enabled dropdown menu on the list entries. Fixed the inmemory service to adhere to the api changes.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/68)

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -22,6 +22,7 @@ import { HeaderComponent} from './header/header.component';
 import { LoginComponent} from './login/login.component';
 import { WorkItemDetailComponent } from './work-item/work-item-detail/work-item-detail.component';
 import { WorkItemListComponent } from './work-item/work-item-list/work-item-list.component';
+import { WorkItemListEntryComponent } from './work-item/work-item-list/work-item-list-entry.component';
 import { WorkItemQuickAddComponent } from './work-item/work-item-quick-add/work-item-quick-add.component';
 import { WorkItemSearchComponent } from './work-item/work-item-search/work-item-search.component';
 import { WorkItemService } from './work-item/work-item.service';
@@ -48,6 +49,7 @@ if (process.env.ENV=='inmemory')
     WorkItemDetailComponent,
     WorkItemQuickAddComponent,
     WorkItemListComponent,
+    WorkItemListEntryComponent,
     WorkItemSearchComponent,
 	StatusDrawerComponent
   ],
--- a/src/app/in-memory-data.service.ts
+++ b/src/app/in-memory-data.service.ts
@@ -2,20 +2,12 @@ import { InMemoryDbService } from 'angular2-in-memory-web-api';
 export class InMemoryDataService implements InMemoryDbService {
   createDb() {
     let workitems = [
-      {"fields":{"system.owner":"tmaeder","system.state":"open"},"id": "21", "name": 'Bullet Chart', description: 'Bullet Chart - Conceptual Design',"workItemType":"1","version":0 , status:'To Do', statusCode:0, type: 'story'},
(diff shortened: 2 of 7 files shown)


src/app/in-memory-data.service.ts
```diff
@@ -2,20 +2,12 @@ import { InMemoryDbService } from 'angular2-in-memory-web-api';
...
+
+      {"fields":{"system.assignee":"someUser1","system.creator":"someOtherUser1","system.description":"Some Description 1","system.state":"new","system.title":"Some Title 1"},"id":"1","type":"system.userstory","version":1},
+      {"fields":{"system.assignee":"someUser2","system.creator":"someOtherUser2","system.description":"Some Description 2","system.state":"open","system.title":"Some Title 2"},"id":"2","type":"system.bug","version":1},
+      {"fields":{"system.assignee":"someUser3","system.creator":"someOtherUser3","system.description":"Some Description 3","system.state":"closed","system.title":"Some Title 3"},"id":"3","type":"system.userstory","version":1},
+      {"fields":{"system.assignee":"someUser4","system.creator":"someOtherUser4","system.description":"Some Description 4","system.state":"resolved","system.title":"Some Title 4"},"id":"4","type":"system.bug","version":1},
+      {"fields":{"system.assignee":"someUser5","system.creator":"someOtherUser5","system.description":"Some Description 5","system.state":"rejected","system.title":"Some Title 5"},"id":"5","type":"system.userstory","version":1}
```
joshuawilson: This is good but should be in a separate commit if not new PR.

michaelkleinhenz: I know, was glitching thru :-/

joshuawilson:
Please squash the merge into the commit.

src/app/work-item/work-item.service.ts
```diff
@@ -10,7 +10,7 @@ export class WorkItemService {
   private headers = new Headers({'Content-Type': 'application/json'});
   // private workItemUrl = 'app/workItems';  // URL to web api
   private workItemUrl = process.env.API_URL+'workitems';  // URL to web api
-  // private workItemUrl = 'http://demo.almighty.io/api/workitems';  // URL to web api
+  // private workItemUrl = 'http://demo.api.almighty.io/api/workitems';  // URL to web api
```
joshuawilson: this should go in a new commit

src/app/in-memory-data.service.ts
```diff
@@ -1,7 +1,7 @@
 import { InMemoryDbService } from 'angular2-in-memory-web-api';
 export class InMemoryDataService implements InMemoryDbService {
   createDb() {
-    let workItems = [
+    let workitems = [
```
joshuawilson: what is the thought behind this change?

michaelkleinhenz: That was a tough one to find. It seems that the name of the variable is actually reflected as the URL part. As the real api uses "../workitems/", this has also to be consistent here.

joshuawilson:
Please rebase on latest master.

michaelkleinhenz:
Done.

src/app/app.module.ts
```diff
@@ -48,6 +49,7 @@ if (process.env.ENV=='inmemory')
     WorkItemDetailComponent,
     WorkItemQuickAddComponent,
     WorkItemListComponent,
+    WorkItemListEntryComponent,
```
sanbornsen: @michaelkleinhenz I guess this should go into WorkItemListComponent instead of main module. Because it's a sub component of WorkItemListComponent. If we keep adding all the sub components here, this list will be crazy big.

michaelkleinhenz:
@sanbornsen mmh..I have deliberately placed it there in fact. I think a central place for registering all used components is desirable. But I also can understand your concern.

PR #94, opened 2016-09-22
Removed the extra assets in the public directory.

michaelkleinhenz:
Removed the public directory with the duped images and the dependency for the unused scss file in the module.

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/94)


joshuawilson:
The idea behind this folder is to hold the static files that do not change. I don't know if we are using it yet. But I would like to see if we can make it work with the build system before we remove them.  

I think that the angular-cli uses this setup.

michaelkleinhenz:
Ok. We should only decide at that time if we use the "assets" or the "public" directory. From the source directory layout, my inner watchdog tends to the asset directory.

PR #106, opened 2016-09-27
Unit test cico builder

sanbornsen:
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/106)

### DIFF
--- a/cico_build_deploy.sh
+++ b/cico_build_deploy.sh
@@ -15,6 +15,16 @@ mkdir -p dist && docker run --detach=true --name=almighty-ui-builder -e "API_URL
 # Build almigty-ui
 docker exec almighty-ui-builder npm install
 
+## Exec unit tests
+docker exec almighty-ui-builder ./run_unit_tests.sh
+
+if [ $? -eq 0 ]; then
+  echo 'CICO: unit tests OK'
+else
+  echo 'CICO: unit tests FAIL'
+  exit 1
+fi
+
 ## Exec functional tests
 docker exec almighty-ui-builder ./run_functional_tests.sh
 
--- a/cico_run_tests.sh
+++ b/cico_run_tests.sh
@@ -17,6 +17,16 @@ mkdir -p dist && docker run --detach=true --name=almighty-ui-builder -e "API_URL
 # Build almighty-ui
 docker exec almighty-ui-builder npm install
 
+## Exec unit tests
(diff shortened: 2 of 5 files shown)


sanbornsen:
@michaelkleinhenz @joshuawilson @nimishamukherjee @SMahil @naina-verma have a look.

cico_build_deploy.sh
```diff
@@ -15,6 +15,16 @@ mkdir -p dist && docker run --detach=true --name=almighty-ui-builder -e "API_URL
 # Build almigty-ui
 docker exec almighty-ui-builder npm install
 
+## Exec unit tests
+docker exec almighty-ui-builder ./run_unit_tests.sh
+
+if [ $? -eq 0 ]; then
```
baijum: Instead of checking like this enable "exit on error" option by setting `set -e`. The `set -x` also can be added.

See this example for core: https://github.com/almighty/almighty-core/blob/master/cico_build_deploy.sh

cico_run_tests.sh
```diff
@@ -17,6 +17,16 @@ mkdir -p dist && docker run --detach=true --name=almighty-ui-builder -e "API_URL
 # Build almighty-ui
 docker exec almighty-ui-builder npm install
 
+## Exec unit tests
+docker exec almighty-ui-builder ./run_unit_tests.sh
+
+if [ $? -eq 0 ]; then
```
baijum: Instead of checking like this enable "exit on error" option by setting `set -e`. The `set -x` also can be added.

See this example for core: https://github.com/almighty/almighty-core/blob/master/cico_run_tests.sh

run_unit_tests.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+LOGFILE=$(pwd)/unit_tests.log
+echo Using logfile $LOGFILE
+
+# Running npm test
+echo Running unit tests...
+npm test >> $LOGFILE
+
+UNIT_TEST_RESULT=$?
```
baijum: Why not display the messages in the standard output? This will appear in Jenkins, right?

ppitonak:
Jenkins job failed

```
exec: "./run_unit_tests.sh": permission denied
CICO: unit tests FAIL
```

You need to set executable flag on script run_unit_tests.sh.

sanbornsen:
Thanks @ppitonak

Issue #107, opened 2016-09-27
Only white spaces for title should not activate the add button in quick add component.


sanbornsen:
This is issue is tested and failed in in this PR - https://github.com/almighty/almighty-ui/pull/98

sanbornsen:
This issue is fixed in this PR - https://github.com/almighty/almighty-ui/pull/108

PR #108, opened 2016-09-27
Check title for only white space in quick add component

sanbornsen:
* Had to bring back the checkTitle function and the flag
* Angular's default keyupevent handler has slight delay and
  that was making the add button visible for a moment.
* This solution works good.


sanbornsen:
This is a fix of issue #107 

@michaelkleinhenz @nimishamukherjee @joshuawilson @SMahil please review the code.

sanbornsen:
@michaelkleinhenz it's failing two of your test cases. But I tried them manually and seems like working fine. Can you have a look?

Expected 'No description available for this work item.' to be 'Some Description'
Expected '' to be 'Some Title'

from - src/tests/work-item/work-item-list/work-item-list.spec.js

PR #114, opened 2016-09-28
No value should contain empty string as title so changing the test case

sanbornsen:


sanbornsen:
This will also fail at jenkins, but this needs to be merged too!

sanbornsen:
This is merged with https://github.com/almighty/almighty-ui/pull/113

PR #115, opened 2016-09-28
Updating work item service create method to accept both inmemory and …

sanbornsen:
…dev env


sanbornsen:
This is merged with https://github.com/almighty/almighty-ui/pull/113

PR #116, opened 2016-09-28
This is functional test file exclusively for MAC El Capitano users to…

sanbornsen:
… run the test locally

### DIFF
--- a/run_functional_tests.sh
+++ b/run_functional_tests.sh
@@ -3,6 +3,16 @@
 LOGFILE=$(pwd)/functional_tests.log
 echo Using logfile $LOGFILE 
 
+# Default value is netcat for linux
+# For MAC the value should be nc
+nc_cmd='netcat'
+is_mac=false
+if [ "$(uname)" == "Darwin" ]; then
+  echo "MAC OS Found"
+  nc_cmd='nc'
+  is_mac=true
+fi
+
 # For the functional tests, we are mocking the core
 export NODE_ENV=inmemory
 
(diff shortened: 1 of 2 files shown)


michaelkleinhenz:
I would strongly recommend that we add your patches for MacOS into the main test runner script. We can't say on what arch the CI is running and that kind of check should be implicit in our process.

I see that you changed ncat to nc. If you want to check if a command is available, you can use bash scripting like this:

NCCOMMAND=nc
if hash ncat 2>/dev/null; then
   NCCOMMAND=ncat
fi
$NCCOMMAND ...params...

More info on how to check if a command is available here:
http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

sanbornsen:
@michaelkleinhenz Yeah, I could do that. **fuser** is there on mac as well but doesn't behave same. And this is a script we are not going to update daily basis. So I thought of creating a separate one instead hacking the main script. A script for mac is not important so I don't want people get confused in future around it. Also, @baijum gave a good point, that shell scripts are mean to be dumb, so we should put as less logic as we can in these script.

michaelkleinhenz:
I think the shell script should run on all arches and the changes are not that big to get it running on both systems. If we do changes to the process we only want to have them implemented in one place..

sanbornsen:
test

sanbornsen:
I don't think we should keep configuration for MAC, not only as conditional in the same file not also as a separate file. The reason behind is:
* There are tons of other Operating Systems out there we can't support all of them.
* This script is to support our build process and shell scripts are mean to be dumb, so we should not put a lot of conditions in it.

I am closing this PR assuming anyone using a separate OS other than Fedora will create their own script.

PR #119, opened 2016-09-29
Fix for #110 - Save works with a blank title on the work item's detailed view page

### LINKED ISSUE #110: Bug - Save works with a blank title on the work item's detailed view page
1. View the details of a work item
2. Clear the title
3. Save
4. Work item gets saved without a title

Proposed fix - save gets disabled if work item title is blank.

nimishamukherjee:
Fix #110.
Previous behaviour - User could click on Save after clearing the title.
Updated behaviour - If the title is cleared:
- Save gets disabled
- An error message is displayed below the Title text field.

### DIFF
--- a/src/app/work-item/work-item-detail/work-item-detail.component.html
+++ b/src/app/work-item/work-item-detail/work-item-detail.component.html
@@ -10,10 +10,12 @@
         {{workItem.id}}
       </div>
     </div>
-    <div class="form-group">
+    <div class="form-group" [ngClass]="{'has-error': !workItem.fields['system.title']}">
       <label class="col-sm-1 control-label" for="wi-detail-title">Title</label>
       <div class="col-sm-5">
         <input type="text" [(ngModel)]="workItem.fields['system.title']" id="wi-detail-title" placeholder="Title (mandatory)" required class="form-control">
+        <span *ngIf="!workItem.fields['system.title']" class="help-block">Title is mandatory.</span>
+
       </div>
     </div>
     <div class="form-group">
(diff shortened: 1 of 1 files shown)


src/app/work-item/work-item-detail/work-item-detail.component.html
```diff
@@ -10,10 +10,12 @@
         {{workItem.id}}
       </div>
     </div>
-    <div class="form-group">
+    <div class="form-group" [ngClass]="{'has-error': !workItem.fields['system.title']}">
       <label class="col-sm-1 control-label" for="wi-detail-title">Title</label>
       <div class="col-sm-5">
         <input type="text" [(ngModel)]="workItem.fields['system.title']" id="wi-detail-title" placeholder="Title (mandatory)" required class="form-control">
+        <span *ngIf="!workItem.fields['system.title']" class="help-block">Title is mandatory.</span>
```
sanbornsen: Validation should not be added like this. We should probably have directive for validation containing all different types of validation logic and messages.

nimishamukherjee: We need to move to Angular Forms for a comprehensive form solution which would include validations.

sanbornsen: Can we do that on this one? Or, is it a future planning?

ppitonak:
Where can I find test for this?

nimishamukherjee:
@ppitonak test cases under https://github.com/almighty/almighty-ui/pull/133 - Save should be disabled if work item's title has only whitespace

src/app/work-item/work-item-detail/work-item-detail.component.html
joshuawilson: I think we can make it as 2 new task items. @nimishamukherjee can you write them up?

@sanbornsen are you good with this answer? or do still have concerns about this code?

sanbornsen: I am good. We can do that.

PR #146, opened 2016-10-06
fixes #141 - Configured tslint to lint on command and in unit tests.

### LINKED ISSUE #141: Make TS linting better
Currently the TS linting is not configured and prints out a lot of warning when you run the unit tests.

joshuawilson:
Add codelizer support to lint Angular 2. Configured tslint to lint on command and in unit tests.

This uses to sets of rules. The main tslint rules can be found at http://palantir.github.io/tslint/rules/

The codelizer is at https://github.com/mgechev/codelyzer. 

The fixes will be in another PR.


config/webpack.test.js
```diff
@@ -204,7 +218,7 @@ module.exports = {
   tslint: {
     emitErrors: false,
     failOnHint: false,
-    resourcePath: 'src'
+    resourcePath: root
```
sanbornsen: I don't know what does root mean. Is it a keyword, which points to the root directory of the project? If it is so, then why are you linting over node_modules and stuff? Shouldn't we only run linting on the code we write?

package.json
```diff
@@ -31,7 +31,7 @@
     "copy-files": "cpy src/index.html src/favicon.ico dist",
     "docker:run": "webpack-dev-server --inline --progress --port 8080",
     "functests": "npm-run-all -p webdriver-start testserver-start protractor",
-    "lint": "tslint ./src/*",
+    "lint": "tslint -c tslint.json --force src/**/*.ts src/**/**/*.ts src/**/**/**/*.ts src/**/**/**/**/*.ts src/**/**/**/**/**/*.ts",
```
sanbornsen: I see what are you trying to do here, but I believe just **src/**/**.ts** should do the trick.

config/webpack.test.js
sanbornsen: 'root' is deprecated, use 'global'

michaelkleinhenz:
Looks good to me. Maybe we have to tweak some rules, but I think we should try it with this set and see if anyone goes rampage (indenting :-)

nimishamukherjee:
Looks good to me. Do we need linting on CSS?

package.json
sanbornsen: Cool :)

PR #153, opened 2016-10-06
Listentry as component

michaelkleinhenz:
Externalizes the entry from the wi list. Fixes some wi service issues to get that running.

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.html
+++ b/src/app/work-item/work-item-list/work-item-list.component.html
@@ -1,70 +1,18 @@
-<div *ngIf="showDialog">	
-	<pf-dialog
-		(pfDialogClick)="onButtonClick($event)"
-		[dialog]='dialog'	
-	></pf-dialog>
-</div>
 <app-header></app-header>
 <div class="container-fluid screen-min-ht containerPaddT">
 	<div id="workItemList_quickAdd" class="width100 fl">
 		<div class="width100 fl supress-hover" (click)="addWorkItem()">
 			<work-item-quick-add (close)="close($event)"></work-item-quick-add>
 		</div>
-	</div>
-	<div class="list-group list-view-pf width100 fl">
-		<div *ngFor="let workItem of workItems; let counter = index" id="{{'workItemList_OuterWrap_'+counter}}" class="work-item-list-entry">
-			<div class="list-group-item" (click)="onSelect(workItem)"
-					 [class.selected]="workItem === selectedWorkItem">
-				<div class="list-view-pf-actions">
-					<button class="btn btn-default workItemList_ViewItemDetailBtn mobdn" (click)="gotoDetail(workItem)">View Details</button>
-					<button class="btn btn-default delete-button workItemList_deleteListItemBtn mobdn" (click)="deleteWorkItem(workItem); $event.stopPropagation()">Delete</button>
-					<div class="dropdown pull-right dropdown-kebab-pf">
-						<button class="btn btn-link dropdown-toggle workItemList_dropdownKebabBtn" type="button"  data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
-							<span class="fa fa-ellipsis-v"></span>
-						</button>
-						<ul class="dropdown-menu dropdown-menu-right" aria-labelledby="dropdownKebabRight">
-							<li><a href="#" class="workItemList_Delete">Delete</a></li>
-							<li><a href="#" class="workItemList_Backlog">Move to backlog</a></li>
-							<li><a href="#" class="workItemList_ChangeStatus">Change Status</a></li>
-						</ul>
-					</div>
-				</div>
-				<div class="list-view-pf-main-info">
-					<div class="list-view-pf-left mobdn">
-						<pf-dropdown
-						 (change)="onStateUpdate($event)" 
-						 [options]="stateDropdownOptions"
-						 [selected]="workItem.selectedState">
-						</pf-dropdown>	
-					</div>
-					<div class="list-view-pf-left type workItemList_workItemType">
-						<!--<span class='fa' [ngClass]="{'fa-bookmark': workItem.workItemType=='story', 'fa-bug':workItem.workItemType=='bug'}"></span>-->
-						<span class='fa' [ngClass]="{'fa-wrench': workItem.type=='system.bug',
-											'fa-user': workItem.type=='system.userstory',
-											'fa-globe': workItem.type=='system.feature',
-											'fa-bookmark': workItem.type=='system.experience',
-											'fa-eye': workItem.type=='system.experience',
-											'fa-bank': workItem.type=='system.fundamental',
-											'fa-rocket': workItem.type=='system.valueproposition' }">
-						</span>
-							{{workItem.id}}
-					</div>
-					<div class="list-view-pf-body">
-						<div class="list-view-pf-description">
-							<div class="list-group-item-heading workItemList_title width100 truncate">
-								{{workItem.fields['system.title']}}
-							</div>
-							<div class="list-group-item-text workItemList_description mobdn">
-								{{workItem.fields['system.description'] ? workItem.fields['system.description'] : "No description available for this work item."}}
-							</div>
-						</div>
-					</div>
-				</div>
-			</div>
-			<!--<div *ngIf="workItem.isExpanded" class="work-item-details" >-->
-      			<!--{{workItem.name | uppercase}} - {{workItem.description}}-->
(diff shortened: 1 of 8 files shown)


src/app/work-item/work-item-list/work-item-list-entry.component.ts
```diff
@@ -0,0 +1,153 @@
...
+    information purposes. The parent MUST NOT delete the workItem associated. The event is
+    intended for display purposes, like removing the entry element and reloading the list.
+*/
+
+@Component({
+  selector: 'work-item-list-entry',
+  templateUrl: '/work-item-list-entry.component.html',
+  styleUrls: ['/work-item-list-entry.component.scss'],
+})
+export class WorkItemListEntryComponent {
```
joshuawilson: Why did you not use "implements OnInit"? You still have an OnInit.

joshuawilson:
@michaelkleinhenz If you want to make the list-entry an external component then you can not refer to "workItem" and it should be moved to the "shared" dir.

If you don't want it to be a shared component, then I feel that this is an unnecessary abstraction.

src/app/work-item/work-item-list/work-item-list-entry.component.scss
```diff
@@ -0,0 +1,22 @@
+@import "../../../assets/stylesheets/custom";
+
+.type span{
```
joshuawilson: Why did you remove the curly braces from around "span"?

src/app/work-item/work-item-list/work-item-list-entry.component.html
```diff
@@ -0,0 +1,54 @@
...
+        <div class="list-view-pf-left mobdn">
+            <pf-dropdown
+                (change)="onChangeState($event)" 
+                [options]="stateDropdownOptions"
+                [selected]="workItem.selectedState">
+            </pf-dropdown>				
+        </div>
+		<div class="list-view-pf-left type workItemList_workItemType">
+			<span class='fa' [ngClass]="{'fa-wrench': workItem.type=='system.bug',
+                                        'fa-user': workItem.type=='system.userstory',
```
joshuawilson: This indent appears to be too much.

src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -1,70 +1,18 @@
...
 	<div id="workItemList_quickAdd" class="width100 fl">
 		<div class="width100 fl supress-hover" (click)="addWorkItem()">
 			<work-item-quick-add (close)="close($event)"></work-item-quick-add>
 		</div>
-	</div>
-	<div class="list-group list-view-pf width100 fl">
-		<div *ngFor="let workItem of workItems; let counter = index" id="{{'workItemList_OuterWrap_'+counter}}" class="work-item-list-entry">
-			<div class="list-group-item" (click)="onSelect(workItem)"
-					 [class.selected]="workItem === selectedWorkItem">
-				<div class="list-view-pf-actions">
```
joshuawilson: I'm going to guess that this section got missed in a merge. Please add it back in.

joshuawilson: Nevermind, I found it.

src/app/work-item/work-item-list/work-item-list-entry.component.ts
michaelkleinhenz: Missed it while reworking it. Funny thing is, that it seems to be called anyway. Will fix it.

src/app/work-item/work-item-list/work-item-list-entry.component.scss
michaelkleinhenz: ? Don't understand.

src/app/work-item/work-item-list/work-item-list-entry.component.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  isSelected(): boolean {
+    return this.selected;
+  }
+
+  // helpers
+
+  confirmDelete(event: MouseEvent) {
+    console.log("HALLO");
```
aslakknutsen: ?

src/app/work-item/work-item-list/work-item-list-entry.component.ts
```diff
@@ -0,0 +1,153 @@
+import { Component, Input, Output, EventEmitter } from '@angular/core';
+import { Router } from "@angular/router";
+
+import { Logger } from '../../shared/logger.service';
+import { DropdownComponent } from './../../shared-component/dropdown/dropdown.component';
+import { DropdownOption } from './../../shared-component/dropdown/dropdown-option';
+import { Dialog } from '../../shared-component/dialog/dialog';
+import { DialogComponent } from '../../shared-component/dialog/dialog.component';
```
michaelkleinhenz: Done.

src/app/work-item/work-item-list/work-item-list-entry.component.ts
michaelkleinhenz: Good find. Leftover.

michaelkleinhenz:
@joshuawilson moved the list entry to shared-component. This is a specific list entry component for displaying work items. It is shared, so putting it in to shared-components makes sense. Defining a generic list entry does not make sense, because that would just be a generic component without any content (it could make sense later on when we need to introduce drag-and-drop, but not yet). So the component is shared, but displays a specific data entity and ensures that work items in lists are looking the same everywhere, have the same controls etc...in work list, in personal lists, in sprint lists etc...

src/app/work-item/work-item-list/work-item-list-entry.component.html
michaelkleinhenz: Fixed. It looks differently in my IDE. I think this is caused by tab/space behaviour.

michaelkleinhenz: Mmmh..checked my vscode config. It seems vscode uses a dynamic configuration here. Tabs are replaced by 2 spaces by default, but when the opened file uses something else (real tabs or different indenting) it uses that instead.

src/app/work-item/work-item-list/work-item-list-entry.component.scss
joshuawilson: It's a SASS thing. In the original version "span" has a set of curly braces around it. In your version they were removed.

src/app/work-item/work-item-list/work-item-list-entry.component.html
joshuawilson: Were you able to fix how vscode works?

joshuawilson:
@michaelkleinhenz Who are you going to share this with? If we have another list view in this app then it will not be a duplicate of the planner. The only reason I can see to break the list-entry out is so that it can be reused. If not it is an extra layer of abstraction that we do not need. It only makes the code harder to maintain.

michaelkleinhenz:
A few motivations:

- the code gets more readable. The combined list + entries was a bit of a mess. It also mixed up the list operations with the entry operations.
- we will re-use the entry component on different views. The work-list is only the work-list (and might eventually become the backlog). This is currently engineered as a "page" or "view", not a component (for example, the quick add is in there). If we see the list as an atomic component, then the quick add should be removed to an enclosing component, because we will eventually having WI lists somewhere where we don't want the quick add (like "search results"). 
- We will need the list entry as an independent component as soon as the drag-and-drop will be introduced. At this point, the code of a combined component will get very complex.
- a list and a listentry are seperate things, should never be in a single entity, my inner Fowler says :-)

src/app/work-item/work-item-list/work-item-list-entry.component.scss
michaelkleinhenz: I did not remove that intentionally. But looking at the code, I would think someone or something has optimized the SASS code. This:

```
.type {
  span{
    ...
  }
}
```

is the semantically same as the version in the PR, the PR version is just optimized. The history says it was never optimized. I don't know where this came from, maybe the old, original CSS or some build has actively changed the file. 
Anyway, the optimized version is imho the better one anyway.

joshuawilson: I was only asking why. I know they are equivalent. I didn't know if @SMahil had done it that way for a reason or not.

joshuawilson:
@michaelkleinhenz ok, if you want to keep it focused on a workItem then please move it under workItem but not under workItemList. It is not truly a shared component as it requires the workItem to function. This makes it a part of the overall workItem architecture.

michaelkleinhenz:
Ok, will do that tomorrow.

PR #154, opened 2016-10-07
Changed tsconfig to comply to new ts format (removed types).

michaelkleinhenz:
Fixes vscode warnings on that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "prebuild": "npm run clean-and-copy",
     "prebuild:prod": "npm run clean-and-copy",
     "protractor": "sleep 15 && ./node_modules/protractor/bin/protractor protractor.config.js",
-    "setup": "npm install && npm run validate",
+    "setup": "npm install && npm run validate && typings install",
     "start": "webpack-dev-server --inline --progress --host 0.0.0.0 --port 8088",
     "test": "karma start",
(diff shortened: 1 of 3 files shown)


joshuawilson:
@michaelkleinhenz This was added after we found out that some of the team was unable to install.  

@nimishamukherjee @sanbornsen @SMahil @naina-verma Can you please test this by doing the following: 
rm -rf node_modules 
rm -rf typings
npm cache clean
npm install

michaelkleinhenz:
Mmmh..install works for me, but not running. It seems it does not install the typings.
I added the typings manually and with --save -G. It created a typings.json. Now everything is running again.

michaelkleinhenz:
@all `typings install` is also required after `npm install`.

joshuawilson:
@michaelkleinhenz The change TS 2 allows us to not need to install the typings anymore. So if you do that you are reverting back to TS1 way of things. Unless you have seen new guidance for TS2.

michaelkleinhenz:
Ah, this gives a hint..

michaelkleinhenz:
Found it. The missing info for me was that the typings was removed due to the new version. The vscode warnings are gone after a cleanroom npm install. Problems were caused by the old ts still hanging around.

PR #160, opened 2016-10-11
Fixes #159 - Update all the TS code to comply with linter

### LINKED ISSUE #159: Fix lint warnings
The new lint checker has found a lot of things and they get listed during the unit test run. They should be cleaned up so that any new code can be written according to the standard.

joshuawilson:
Fixes #159 - Update all the TS code to comply with linter.
Update functional tests to work with new list-item-entry component. They were failing.

### DIFF
--- a/src/tests/work-item/work-item-list/work-item-list.spec.js
+++ b/src/tests/work-item/work-item-list/work-item-list.spec.js
@@ -65,7 +65,7 @@ describe('Work item list', function () {
       expect(page.workItemDescription(page.workItemByNumber(0))).toBe('Some Description');
       expect(page.workItemTitle(page.workItemByNumber(0))).toBe('Some Title');
       page.allWorkItems.getText().then(function (text) {
-        expect(text).toContain("View Details Delete\nnew\n1\nSome Title\nSome Description");
+        expect(text).toContain("new\n1\nSome Title 1\nSome Description 1\nView Details Delete");
       });
     });
   });
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,7 +1,7 @@
 import { Component } from '@angular/core';
 
 @Component({
-  selector: 'my-app',
+  selector: 'alm-app',
   templateUrl: './app.component.html',
   styleUrls: ['./app.component.scss']
   })
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -1,31 +1,46 @@
 import './rxjs-extensions';
 
 import { NgModule }      from '@angular/core';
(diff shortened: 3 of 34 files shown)


joshuawilson:
@michaelkleinhenz @ldimaggi Can you please review this PR. The 2nd commit is to fix a problem with the merger of the code you both submitted. I didn't realize @michaelkleinhenz had not rebased to include @ldimaggi new tests. Please check the fix I did to the tests and confirm is it correct.

To run the linter use: npm run lint

src/tests/work-item/work-item-list/work-item-list.spec.js
```diff
@@ -65,7 +65,7 @@ describe('Work item list', function () {
       expect(page.workItemDescription(page.workItemByNumber(0))).toBe('Some Description');
       expect(page.workItemTitle(page.workItemByNumber(0))).toBe('Some Title');
       page.allWorkItems.getText().then(function (text) {
-        expect(text).toContain("View Details Delete\nnew\n1\nSome Title\nSome Description");
+        expect(text).toContain("new\n1\nSome Title 1\nSome Description 1\nView Details Delete");
```
michaelkleinhenz: Works.
Just as a general comment: I am not sure that is the recommended way of doing that. getText() "flattens" the text in the element including all childs. This might work (it's flaky in my opinion), but it essentially checks the hierarchy of the childs. If we decide to put the details button before the delete button for example, this would break.

michaelkleinhenz: Ah, and that was was happening here effectively and was the cause of the error. In the old code, the buttons were defined before the content but displayed after the content. I "ordered" that so that the hierarchy matches the display and it broke. So it not checks the display but the hierarchy, what I think is not the point. I'll talk with Len about that.

michaelkleinhenz:
Works and tests ok. Do we really have a lint rule that requires { } around one line blocks? :-/

src/tests/work-item/work-item-list/work-item-list.spec.js
ldimaggi: Looks good! Please commit!

Yes - the test code was pretty brittle - sorry that it caused a problem! It was actually a work-around to how the UI included the quick add web elements as the first "row" in the display of work items. Is this how we will ensure that the quick add web elements are always displayed at the "top" of the page? Thx!

joshuawilson: @ldimaggi no, we need to find a better way. It was more of just a way to show something until the designs came out.

PR #176, opened 2016-10-13
Fixes #134 - Make it easier to run the test.

### LINKED ISSUE #134: Readme and npm scripts should be updated
Readme file mentions these two commands for running tests

```
./run_unit_tests.sh
./run_functional_tests.sh
```

Standard way to run (unit) tests in Node.js/NPM projects is to run `npm test` which actually exists in almighty-ui but it doesn't work.

There should be unified way how to run tests, e.g. update _test_ script in package.json to call `run_unit_tests.sh` and create something like _test:functional_ in package.json that would call  `run_functional_tests.sh`

joshuawilson:
Fixes #134 - Make it easier to run the test. Update the README with instructions.

Now "npm test" will run every check we have. There are separate commands for each check and they are now explained in the README.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,9 +1,18 @@
 = ALMighty UI
 
+=== ALMighty UI is a currently an application planner and issue tracker front-end.
+It uses https://github.com/almighty/almighty-core[ALMighty-core] as the back-end.
+
 If this is the first time you are starting the UI you need to run
 
 ----
 $ npm install
+----
+
+If you trying to refresh your install you can run:
+
+----
+$ npm run reinstall
 ----
 
 Start the app by executing the following. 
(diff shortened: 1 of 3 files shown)


README.adoc
```diff
@@ -1,40 +1,96 @@
 = ALMighty UI
 
+=== ALMighty UI is a currently an application planner and issue tracker front-end.
+It at uses https://github.com/almighty/almighty-core[ALMighty-core] as the back-end.
```
nimishamukherjee: Remove "at"

README.adoc
```diff
@@ -1,40 +1,96 @@
...
+
 If this is the first time you are starting the UI you need to run
 
 ----
 $ npm install
 ----
 
+If you trying to refresh your install you can just run
```
nimishamukherjee: We can remove "just" and have "you can run"

sanbornsen:
LGTM

README.adoc
```diff
@@ -1,40 +1,96 @@
...
+
 == Unit Tests
 
+To run the unit test, use:
+
+----
+$ npm run test:unit
+----
+
+To run the unit test in a watch mode so that the are rerun every time you save, use:
```
nimishamukherjee: Use "they" instead of "the"
..in a watch mode so that "they"

README.adoc
joshuawilson: thanks

README.adoc
```diff
@@ -1,40 +1,96 @@
...
+
+----
+$ npm watch:test
+----
+
 Unit (UI) tests are automatically executed on the PR check process (using
```
nimishamukherjee: As this is a Readme, we could expand PR (pull request) once and then use the abbreviation for the rest of the document.

joshuawilson: agreed

README.adoc
```diff
@@ -1,40 +1,96 @@
...
-`cico_run_tests.sh`) or on the release process (using `cico_build_deploy.sh`) but
-can also executed from the development environment by invoking the following command:
+`cico_run_tests.sh`) or on the release process (using `cico_build_deploy.sh`). To executed
+these scripts from the development environment you can run the following script:
 
 ----
 ./run_unit_tests.sh
 ----
 
 For unit testing, Jasmine, Karma, Webpack is used. There are unit tests available for
```
nimishamukherjee: For unit testing, Jasmine, Karma, Webpack is used.
Do we use is or are?

README.adoc
```diff
@@ -1,40 +1,96 @@
...
-Functional (UI) tests are automatically executed on the PR check process (using 
+To run the functional test, use:
+
+----
+$ npm run test:func
+----
+
+Functional (UI) tests are automatically executed on the PR check process (using
 `cico_run_tests.sh`) or on the release process (using `cico_build_deploy.sh`) but
 can also executed from the development environment by invoking the following command:
```
nimishamukherjee: can also executed - add "be"

README.adoc
joshuawilson: I think "are" should be used here.

README.adoc
joshuawilson: ok

PR #187, opened 2016-10-14
Unwanted use of event removed.

sanbornsen:


sanbornsen:
This is the fix of - https://github.com/almighty/almighty-ui/issues/177

PR #192, opened 2016-10-16
Using set instead of append to add auth header.

sanbornsen:


sanbornsen:
This is fix of issue - https://github.com/almighty/almighty-ui/issues/191

@nimishamukherjee @joshuawilson @michaelkleinhenz @aslakknutsen @SMahil please review it.

src/app/shared/broadcaster.service.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class Broadcaster {
+  private _eventBus: Subject<BroadcastEvent>;
+
+  constructor() {
+    this._eventBus = new Subject<BroadcastEvent>();
+  }
+
+  broadcast(key: any, data?: any) {
+    console.log("I am here");
```
nimishamukherjee: Remove this console.log :)

nimishamukherjee:
I'm fine with the changes.

michaelkleinhenz:
Looks good. General suggestion: I think we have to do more source comments. That would help to understand what the source is actually supposed to do.

joshuawilson:
[test]

Issue #199, opened 2016-10-18
Logout should not reload the page.


sanbornsen:
Logout should broadcast an event and all the places should be updated listening to that event.

nimishamukherjee:
PR #189 merged with master, hence closing this task.

Issue #198, opened 2016-10-18
Remove token param from the URL after login.


sanbornsen:
After login we have to remove the token from the URL as soon as possible.

sanbornsen:
Done!

PR #200, opened 2016-10-18
Logout event listener to avoid page reloading

sanbornsen:
- An event is broadcasted over logout
    - Listened through out the comonents
    - Update components accordingly on logout
    - Also removing token from the URL after login
            - Need to find a better way to do it, other than just
	      reloading the app to the base url.

### DIFF
--- a/src/app/auth/authentication.service.ts
+++ b/src/app/auth/authentication.service.ts
@@ -23,6 +23,7 @@ export class AuthenticationService {
     if ('token' in params) {
       this.authToken = params['token'];
       localStorage.setItem('auth_token', this.authToken);
+      location.href = location.protocol + '//' + location.host;
       return true;
     }
     //this.router.navigate(['login']);
@@ -32,9 +33,9 @@ export class AuthenticationService {
   logout() {
     this.authToken = '';
     localStorage.removeItem('auth_token');
-    //this.router.navigate(['login']);
+    // this.router.navigate(['login']);
     this.broadcaster.broadcast('logout', 1);
-    location.href = location.protocol + '//' + location.host + location.pathname + location.hash;
+    // location.href = location.protocol + '//' + location.host + location.pathname + location.hash;
   }
 
   getToken() {
--- a/src/app/work-item/work-item-list-entry/work-item-list-entry.component.ts
+++ b/src/app/work-item/work-item-list-entry/work-item-list-entry.component.ts
@@ -2,6 +2,7 @@ import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { Router }                                         from '@angular/router';
 
 import { AuthenticationService } from './../../auth/authentication.service';
+import { Broadcaster } from './../../shared/broadcaster.service';
 import { Logger } from '../../shared/logger.service';
 
 import { Dialog }            from '../../shared-component/dialog/dialog';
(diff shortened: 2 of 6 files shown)


sanbornsen:
This is for the following issues - 
#199 
#198 - Still not the best way

src/app/auth/authentication.service.ts
```diff
@@ -23,6 +23,7 @@ export class AuthenticationService {
     if ('token' in params) {
       this.authToken = params['token'];
       localStorage.setItem('auth_token', this.authToken);
+      location.href = location.protocol + '//' + location.host;
```
sanbornsen: Did this to get rid of the token from the URL. Need to figure out a better way. Need help!!

nimishamukherjee: The token does not come in the URL but the page refresh is there.

We need to consider this scenario - the user needs to be sent back to the page from where he clicked login. 
If the user is on View Details and clicks Login, after the login is done, the View Details screen needs to be displayed. 
We need to store the route from where login was called, and after successful login, the user needs to be sent back to the same page.

@joshuawilson your thoughts?

nimishamukherjee:
[test]

sanbornsen:
[test]

sanbornsen:
I did nothing and it passed just by running once again.

joshuawilson:
@nimishamukherjee please let me know when you have reviewed this.

src/app/work-item/work-item-list-entry/work-item-list-entry.component.ts
```diff
@@ -2,6 +2,7 @@ import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { Router }                                         from '@angular/router';
 
 import { AuthenticationService } from './../../auth/authentication.service';
+import { Broadcaster } from './../../shared/broadcaster.service';
```
joshuawilson: You don't need the ./ in front of the path if you have ../ to begin with.

src/app/work-item/work-item-list-entry/work-item-list-entry.component.ts
```diff
@@ -51,9 +52,14 @@ export class WorkItemListEntryComponent implements OnInit {
   loggedIn: Boolean = false;
 
   constructor(private auth: AuthenticationService,
+              private broadcaster: Broadcaster,
               private router: Router,
               private workItemService: WorkItemService,
               private logger: Logger) {
+                 this.broadcaster.on<string>('logout')
```
joshuawilson: Should this really be done in the constructor. There are several different life cycle methods that you may want to consider, like ngOnInit. You really need to ask yourself when do you want this method called.

joshuawilson: Here is a reference to it in the docs. https://angular.io/docs/ts/latest/guide/lifecycle-hooks.html

src/app/work-item/work-item-list-entry/work-item-list-entry.component.ts
sanbornsen: But we agreed on following one standard.

src/app/work-item/work-item-list-entry/work-item-list-entry.component.ts
sanbornsen: Yeah, agree. Constructor is not the best place for it. I'll do the changes.

nimishamukherjee:
[test]

joshuawilson:
[test]

nimishamukherjee:
[test]

nimishamukherjee:
[test]

joshuawilson:
[test]

joshuawilson:
@sanbornsen It appears that @nimishamukherjee squashed the branch this PR is based on. You will need to try and rebase it against the updated branch. If you can't get it to rebase properly, you may need to create a new branch based on the github-login branch and copy your work over to it.

sanbornsen:
[test]

PR #202, opened 2016-10-18
Only set Authorization header if Token exist

aslakknutsen:
* Call all API's with Authorization header

### DIFF
--- a/src/app/work-item/work-item.service.ts
+++ b/src/app/work-item/work-item.service.ts
@@ -17,7 +17,9 @@ export class WorkItemService {
   constructor(private http: Http,
               private logger: Logger,
               private auth: AuthenticationService) {    
-    this.headers.append('Authorization', 'Bearer ' + this.auth.getToken());
+    if(this.auth.getToken() != null) {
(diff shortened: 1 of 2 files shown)


nimishamukherjee:
[test]

nimishamukherjee:
[test]

src/app/work-item/work-item.service.ts
```diff
@@ -17,7 +17,9 @@ export class WorkItemService {
   constructor(private http: Http,
               private logger: Logger,
               private auth: AuthenticationService) {    
-    this.headers.append('Authorization', 'Bearer ' + this.auth.getToken());
+    if(this.auth.getToken() != null) {
```
aslakknutsen: maybe this should call this.auth.isLoffedIn() instead of checking token status?

joshuawilson: @aslakknutsen do you want to change it before I merge it?

joshuawilson: @aslakknutsen I merged it, if you want to make that change please send a new PR once this branch is in master.

PR #219, opened 2016-10-20
Able to add Work Items in Mobile List View

SMahil:


joshuawilson:
@SMahil I get the following warning in the browser console when I run this.

The value "device-width;" for key "width" is invalid, and has been ignored.
The value "1.0;" for key "initial-scale" was truncated to its numeric prefix.
The value "1.0;" for key "maximum-scale" was truncated to its numeric prefix.
The value "0;" for key "user-scalable" was truncated to its numeric prefix.
Error parsing a meta element's content: ';' is not a valid key-value pair separator. Please use ',' instead.

PR #221, opened 2016-10-21
Added page object functions for new quick add layout 

naina-verma:
Added page object functions for new quick add layout 
- Added Handles for Confirm delete/cancel button
- Click on "Add quick workitem" as per new layout

- New spec file to test quick add related tests (quickadd-workitem.spec.js)

### DIFF
--- a/src/tests/work-item/work-item-list/quickAdd-workitems.spec.js
+++ b/src/tests/work-item/work-item-list/quickAdd-workitems.spec.js
@@ -1,0 +1,82 @@
+/**
+ * POC test for automated UI tests for ALMighty
+ *@auther Naina
+ * Note on screen resolutions - See: http://www.itunesextractor.com/iphone-ipad-resolution.html
+ * Tests will be run on these resolutions:
+ * - iPhone6s - 375x667
+ * - iPad air - 768x1024
+ * - Desktop -  1920x1080
+ *
+ * beforeEach will set the mode to phone. Any tests requiring a different resolution will must set explicitly.
+ *
+ */
+
+var WorkItemListPage = require('./work-item-list.page');
+
+describe('Work item list', function () {
+  var page, items, startCount, browserMode;
+
+  beforeEach(function () {
+    setBrowserMode('phone');
+    page = new WorkItemListPage();
+    page.allWorkItems.count().then(function(originalCount) { startCount = originalCount; });
+  });
+  it('Creating a new quick add work item and delete - phone.', function () {
+    setBrowserMode('phone');
+    page.clickQuickAddWorkItemTitleButton();
+    page.typeQuickAddWorkItemTitleText('Quick Add and Delete');
+    page.clickWorkItemQuickAddButton().then(function() {
+    expect(page.workItemTitle(page.firstWorkItem)).toBe('Quick Add and Delete');
+    expect(page.workItemTitle(page.workItemByNumber(0))).toBe('Quick Add and Delete');
+    page.clickWorkItemKebabButton();
+    page.clickWorkItemKebabDeleteButton();
+    page.clickWorkItemPopUpDeleteConfirmButton().then(function() {
+    expect(page.workItemTitle(page.firstWorkItem)).not.toBe('Quick Add and Delete');
+    expect(page.workItemTitle(page.workItemByNumber(0))).not.toBe('Quick Add and Delete');
+    });
+    });
+
+
+  });
+
+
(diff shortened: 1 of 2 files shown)


ldimaggi:
In running the tests locally, the first problem that I see is that when we remove the browser.get statement from WorkItemPage definition:

```
  browser.get("http://localhost:8088/");
};
```
The tests that create a new workitem and access a workitem's detail page fail without the creation of new browser in each test. What was the main reason to remove the statement? It might be less efficient to create a new page object for each test, but it does guarantee that each test "starts clean" with the browser in a known state. Thx!

ldimaggi:
After adding back the statement to create the new browser, the tests all run, but I am seeing these warnings - looking now for how to clear these - the tests should always reference precise UI elements:

.[11:53:11] W/element - more than one element found for locator By(css selector, *[id="dropdownKebabRight"]) - the first result will be used
[11:53:12] W/element - more than one element found for locator By(css selector, .workItemList_Delete) - the first result will be used
[11:53:12] W/element - more than one element found for locator By(css selector, .btn.btn-primary.alm-dialog-btn.float-left) - the first result will be used
.[11:53:20] W/element - more than one element found for locator By(css selector, *[id="dropdownKebabRight"]) - the first result will be used
[11:53:20] W/element - more than one element found for locator By(css selector, .workItemList_Delete) - the first result will be used

naina-verma:
@ldimaggi , I have added 
  browser.get("http://localhost:8088/");
};

joshuawilson:
@naina-verma also, can you please squash these commits into 1 commit. (unless you have a compelling reason to keep them separate)

ldimaggi:
One important thing for any/every UI test is that the test is able to access the precise UI element. It's sometimes a pain to write the tests in this way, but it is important that we are able to be 100% certain that the test is accessing the UI correct UI elements.

The key to finding the correct dropdownKebabRight element is to find it relative to a specific work item. I've included a working example that be added to one of the test spec files (see below). I have intentionally not created a new branch and submitted a pull request for this as you already have a pull request in progress.

The key is the use of the workItemTitle function in the WorkItemListPage module as this function is written to return a work item title element that is found relative to a work item UI element.

I tried many different approaches to enclose this code in a callable function, but given the requirement to use nested promises, I was not able to get that to work. (Maybe Josh or Michael know of some magic to work around this. ;-)

Anyway, all you have to do to access a work item's dropdownKebabRight UI element is to create a function similar to
WorkItemListPage.workItemTitle().

Have a great weekend!


``

     page.workItemTitle(element).then(function(text) {
        if (testWorkItemTitle == text) {     
          var returnedWorkItem = page.workItemByIndex(index);
          page.workItemTitle(returnedWorkItem).then(function (text) { 
            console.log("We found the work item with title " + text + " - now run a test...");
            /* Access elements relative to the workitem and test here */             
          }); /* then for work item title  console.log statement */  
        }  /* if - was the intended work item found */
      }); /* then for the text of the work item */    
    }); /* each for all the elements in the UI */     
  }); /* it */

ppitonak:
@joshuawilson can you whitelist Naina so that Jenkins jobs run automatically?

joshuawilson:
closing as it is now out of scope

ldimaggi:
Here;s an example of what I suggested regarding accessing UI elements relative to other UI elements:


**In the page object:**

```
   newClickWorkItemKebabButton:  {
    value: function (workItemElement)
    { return workItemElement.element(by.id("dropdownKebabRight")).click(); }
  },
  
  newClickWorkItemKebabDeleteButton:   {
    value: function (workItemElement)
    { return workItemElement.element(by.css('.workItemList_Delete')).click(); }
  },

```

**And here's a simple test:**

```
  it('delete a workitem - phone.', function () {
    setBrowserMode('phone');   
    var testWorkItemTitle = "Some Title 14";
    
    /* Access elements relative to the workitem and test here */   
    var returnedWorkItem = page.firstWorkItem;  
    page.newClickWorkItemKebabButton(returnedWorkItem);     
    page.newClickWorkItemKebabDeleteButton(returnedWorkItem);
    page.clickWorkItemPopUpDeleteConfirmButton().then(function() {
      expect(page.workItemTitle(page.firstWorkItem)).not.toBe(testWorkItemTitle);
      expect(page.workItemTitle(page.workItemByNumber(0))).not.toBe(testWorkItemTitle);
    });              
  });

```

These changes will get rid of the warnings caused by duplicate UI elements being found.

PR #232, opened 2016-10-24
Adding page object functions for new quick add layout

naina-verma:
Added page object functions for new quick add layout

Added Handles for Confirm delete/cancel button
Click on "Add quick workitem" as per new layout

New spec file to test quick add related tests (quickadd-workitem.spec.js)


joshuawilson:
closing as it has the wrong commit

PR #235, opened 2016-10-25
Fixes #175 - Update application to use Angular 2.1.1

### LINKED ISSUE #175: Upgrade to Angular 2.1.0

joshuawilson:
DO NOT MERGE. Tests fail. Please review and help me fix the failing tests.

Additionally, the inmemory loading in the module failed and I had to change it. If anyone can find a more elegant way fix this, please let me know.

NOTE: you will need to run 'npm run reinstall'


nimishamukherjee:
@joshuawilson this PR fails on my local machine.
I did a npm run reinstall + npm test

* npm test fails
* npm run test:unit passes
* npm run test:func fails

functionaltestfail1
functionaltestfail2

src/app/work-item/work-item-detail/work-item-detail.component.ts
```diff
@@ -1,6 +1,6 @@
-import { Location }               from '@angular/common';
 import { Component, OnInit }      from '@angular/core';
 import { ActivatedRoute, Params } from '@angular/router';
+import { Location }               from '@angular/common';
```
ppitonak: This change probably shouldn't be here.

package.json
```diff
@@ -87,11 +88,11 @@
     "istanbul-instrumenter-loader": "0.2.0",
     "jasmine-core": "2.4.1",
     "karma": "1.3.0",
-    "karma-chrome-launcher": "0.2.3",
-    "karma-cli": "0.1.2",
+    "karma-chrome-launcher": "2.0.0",
```
ppitonak: Is upgrade of Karma and Protractor necessary in this PR?

michaelkleinhenz:
I am collecting findings here:
* There seems to be some problem with Protractor 4.0.9, if I revert to 4.0.8, I get different errors. Suspicious is that I then get fails on the exact same 3 tests as in the build log here. If I use 4.0.9, all tests fail with the "undefined is not a function (evaluating 'window.getAllAngularTestabilities()" error.
* In the test log, there are several SASS errors that point towards some incompatibilities with the build system: "Module build failed: Error: Node Sass does not yet support your current environment: Linux 64-bit with Unsupported runtime (51) For more information on which environments are supported please see: https://github.com/sass/node-sass/releases/tag/v3.10.1". 
* The failing SASS compiler results in errors all over the place, also while the tests are running. From the errors, I suspect that the polyfills try to load a CSS file, which fails, and then times out, resulting in the test failures. That would also explain why only a few tests fail: only the tests requiring a loading of CSS resources inside the polyfills fail.

michaelkleinhenz:
Sidenote: in the page model files, there are error messages defined that are supposed to be logged when an element can not be found after a timeout, but that messages are not appearing in the log or on the console. @ldimaggi

michaelkleinhenz:
Indeed, the test is failing. On the first test for example, the call to "clickQuickAddWorkItemTitleButton()" fails with a timeout waiting for a button with "Cancel" appearing.

michaelkleinhenz:
Found it: the problem is that the event propagation on the "Delete" link is not working. The behaviour is as follows: if I click on the "Delete" link in an item's dropdown, the confirmation dialog appears briefly, but then the page gets reloaded. The test then waits until the timeout for the "Cancel" button to appear.

Proof: if I add `event.preventDefault()` in the `confirmDelete` method in the `work-item-list-entry.component.ts` (line 93):

```
confirmDelete(event: MouseEvent) {
    event.stopPropagation();
    **event.preventDefault();**
    this.dialog = {
```

It works manually and the tests are running smooth.

michaelkleinhenz:
Addition: @sanbornsen suggested removing the href="#" as an alternative. Also works to prevent events propagated along the chain.

src/app/work-item/work-item-detail/work-item-detail.component.ts
joshuawilson: These are changes that the Angular team did in the docs.

package.json
joshuawilson: These are changes that the Angular team did in the docs.
Though we found an issue with Protractor, so I downgraded it to 4.0.8.

joshuawilson:
[test]

sanbornsen:
[test]

sanbornsen:
[test]

PR #239, opened 2016-10-26
remove auto-upgrade for protractor and print functional logs

joshuawilson:


joshuawilson:
this does not fix the issue

PR #245, opened 2016-10-26
Add optional description field in Quick Add Work Item

SMahil:


sanbornsen:
@joshuawilson @SMahil 
Quick add component looks broke - http://imgur.com/a/veErv

PR #247, opened 2016-10-27
set protractor to stay at 4.0.8

joshuawilson:
also reset to early commit


joshuawilson:
testing resetting, does not work.

PR #248, opened 2016-10-27
Moved function that sets browser window size to test support file so …

ldimaggi:
…tests will not have to duplicate the function. Added a new (empty) template file to assist writers in adding tests.

### DIFF
--- a/src/tests/work-item/work-item-list/test-template.spec.js
+++ b/src/tests/work-item/work-item-list/test-template.spec.js
@@ -1,0 +1,30 @@
+/**
+ * Template for automated UI tests for ALMighty
+ *  
+ * Note on screen resolutions - See: http://www.itunesextractor.com/iphone-ipad-resolution.html
+ * Tests will be run on these resolutions:
+ * - iPhone6s - 375x667
+ * - iPad air - 768x1024
+ * - Desktop -  1920x1080
+ * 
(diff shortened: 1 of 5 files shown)


sanbornsen:
Neat !! @ldimaggi You need to configure your editor for indentation!!

src/tests/work-item/work-item-list/test-template.spec.js
```diff
@@ -0,0 +1,30 @@
...
+describe('Test name here...', function () {
+  var page, items, browserMode;
+
+  beforeEach(function () {
+    testSupport.setBrowserMode('phone');
+    page = new WorkItemListPage();    
+  });
+
+  it('should testsomething here...', function() {
+//    expect(something...).toBe(comething...);
```
joshuawilson: is this an example or are you missing a test? a placeholder to be implemented later?

joshuawilson: nevermind, I get it. very nice :)

ldimaggi: Only a placeholder - thx!

joshuawilson:
@ldimaggi add to whitelist

joshuawilson:
@ldimaggi the last comment is supposed to add you to the list so that the system will auto run the ci server checks.

PR #252, opened 2016-10-29
Refectored page objkect to be easier to use and maintain. Added new test.

ldimaggi:
This should make the page objects easier to maintain and more reliable.


joshuawilson:
Why is there a number in one of the file names?

ldimaggi:
That's the user story number.


(Sent from my iPhone)



On Oct 30, 2016, at 1:27 AM, Joshua Wilson <notifications@github.com> wrote:

Why is there a number in one of the file names?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/almighty/almighty-ui/pull/252#issuecomment-257132962>,
or mute the thread

joshuawilson:
but why is it in the file name? Isn't that a bit specific and how would anyone else tie that number back to the user story?

michaelkleinhenz:
Looks fine. I also would not use the task id in filenames.

ldimaggi:
Agreed - just changed the file name. Thx!

PR #257, opened 2016-11-01
Re-arranging the app following NgModule structure

sanbornsen:
- Work item module added.
- Router for work item added.


src/app/login/login.component.html
```diff
@@ -10,13 +10,11 @@
     </div><!--/.col-*-->
 
     <div class="col-sm-7 col-md-6 col-lg-5 login">
-      <form  id="login_githubLoginForm" class="form-horizontal" role="form">
```
joshuawilson: Why did you take the Form out of this?

sanbornsen: Without the FormsModule directive, that button in the form behaves like a submit button. So it just reloads the page as a submission behavior. FormsModule is not necessary for the AppModule as a root import if we simply don't use the form tag in login html. If we still want to have it in the HTML, then we can make login as a Module and import it there, which sounds reasonable, because anyway we are going to make login page modular.

PR #262, opened 2016-11-02
Using Ng2Dropdown for kebob menu

sanbornsen:


src/app/work-item/work-item-list/work-item-list.module.ts
```diff
@@ -12,6 +13,7 @@ import { WorkItemListEntryComponent } from './work-item-list-entry/work-item-lis
   imports:      [
     CommonModule, 
     WorkItemQuickAddModule,
+    Ng2Dropdown,
     DropdownModule,
```
joshuawilson: Do you still need the DropdownModule if you replaced it with a new name?

If you need the old name then don't make a new one until you can change both.

sanbornsen: This DropdownModule the one I created, and Ng2Dropdown uses the same name as DropdownModule. For now I need both of them, so renamed one while importing up there.

joshuawilson: ok, I see it now

PR #264, opened 2016-11-02
Fixing quick add bugs

sanbornsen:


sanbornsen:
This PR is to fix - https://github.com/almighty/almighty-ui/issues/259

src/app/work-item/work-item-quick-add/work-item-quick-add.component.html
```diff
@@ -11,19 +11,24 @@
     </div>
     <div class=" workItemQuickAdd_story">
       <div class="col-md-6 col-sm-6 col-xs-12 workItemQuickAdd_storyInput">
-        <input #quickAdd type="text" id="exampleInput" class="form-control width100 pull-left"  [(ngModel)]="workItem.fields['system.title']" 
-              (keyup.enter)="save()"
+        <input #quickAddTitle type="text" id="exampleInput" class="form-control width100 pull-left"  [(ngModel)]="workItem.fields['system.title']" 
+              (keydown.enter)="save()"
```
joshuawilson: why change it to keydown? that means that if you press and hold enter it fires the save

src/app/work-item/work-item-quick-add/work-item-quick-add.component.html
```diff
@@ -11,19 +11,24 @@
...
               (keyup)="checkTitle()" placeholder="A New Story" required/>
       </div>
       <div class="col-md-6 col-sm-6 mobdn">
-        <textarea type="text" rows="1" id="exampleDesc" class="form-control width100 pull-left"  [(ngModel)]="workItem.fields['system.description']" 
-               placeholder="Description">
+        <textarea #quickAddDesc rows="1" type="text" id="exampleDesc" class="form-control width100 pull-left"  [(ngModel)]="workItem.fields['system.description']" 
+                [style.height]="descHeight"
+                [style.resize]="descResize"
+                (keydown.enter)="save()"
```
joshuawilson: Why do we need both keydown and keyup. I would think it would fire twice.

src/app/work-item/work-item-quick-add/work-item-quick-add.component.html
sanbornsen: By keydown I fire the event as soon as the key is pressed. You don't have to hold it. In case of keyup, if you press and then leave the key, then the event get fired.

src/app/work-item/work-item-quick-add/work-item-quick-add.component.html
sanbornsen: In case of a text area we need both the event so that, none of this behaves like an Enter and create a line break. 

- Event on keydown - 
      - You press enter it saves and clean the field
      - You leave the enter it creates a line break in the field
- Event on keyup - 
      - You press the enter it creates a line break 
      - You leave it, it saves the text with the line break
- Event on both - 
      - You press the enter it triggers save and reset the field
      - You leave the key it triggers save again with an empty field which basically does nothing.

PR #273, opened 2016-11-08
User github Login automation 

naina-verma:
As a end user of ALM , we need to login to alm using github account .
- Added  Page object model for github login Mobile/Desktop
- Added login.spec.js with two login tests with mobile & desktop (For upcoming tests related login too)

### DIFF
--- a/src/tests/work-item/work-item-list/login.spec.js
+++ b/src/tests/work-item/work-item-list/login.spec.js
@@ -1,0 +1,62 @@
+/**
+ * POC test for automated UI tests for ALMighty
+ *  Story: Display and Update Work Item Details
+ *  https://github.com/almighty/almighty-core/issues/296
+ *
+ * Note on screen resolutions - See: http://www.itunesextractor.com/iphone-ipad-resolution.html
+ * Tests will be run on these resolutions:
+ * - iPhone6s - 375x667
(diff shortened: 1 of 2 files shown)


joshuawilson:
@naina-verma this doesn't pass for me when I run it locally.

naina-verma:
@joshuawilson Hi,

I have checked it again , the problem is not with Test cases , the problem is:
- When we click on Sign in with github it does not redirect user to gitbug signin page ,instead it goes to localhost:8088/app/login
You can put this PR on hold i will figure out the solution and update

ldimaggi:
The problem with the test is that by default, the tests access the workitems' UI on localhost at this URL:
 browser.get("http://localhost:8088/?token=justarandomtokenfortest

Redirection to github is not working for login when the tests/server is localhost. I was able to run the login test after making the following code change - but all the other tests (as expected  ;-) then fail.

```
//  browser.get("http://localhost:8088/?token=justarandomtokenfortest");
  browser.get("http://demo.almighty.io");
};
```

joshuawilson:
@naina-verma do you plan to fix this or should we close it?

naina-verma:
@joshuawilson , am closing this PR it needs more R&D and changes in code. I will open it later with new changes. 
Thanks

PR #287, opened 2016-11-09
Outline around description field is take care of. (fixes #278)

### LINKED ISSUE #278: Outline Around Description Field
According to the [Visual Design](https://redhat.invisionapp.com/share/ZE991DOJ6#/195986245_A-0400-Details_Panel), there should not be an outline of the Description field, unless it is in the active/edit state.

descriptionoutline

sanbornsen:
- User hovers over the description.
- A border with that edit icon appears.
- Clicking on the box or the edit icon would make the border highlighted
with light blue.
- And user can edit.
- User can press ESC while editing to cancel or Enter to save.
- Pressing ESC twice will close the panel.


nimishamukherjee:
On hover the text has a "jumping" effect (on Chrome)
Can the padding/margin be set so that the border appears without the text moving.

PR #296, opened 2016-11-10
Updating detail viewcases

naina-verma:
Added test cases For detail page  …
- Edit with blank title - Mobile/Desktop Should show validation message
- Edit title and description by hitting Enter key Mobile/Desktop

### DIFF
--- a/src/tests/work-item/work-item-list/displayAndUpdateWorkItemDetails.spec.js
+++ b/src/tests/work-item/work-item-list/displayAndUpdateWorkItemDetails.spec.js
@@ -85,7 +85,6 @@ var waitTime = 30000;
         //detailPage.clickWorkItemTitleEditIcon();
         detailPage.setWorkItemDetailTitle (workItemUpdatedTitle, false);
         detailPage.clickWorkItemTitleSaveIcon();
-
 //        detailPage.clickWorkItemDescriptionEditIcon();
         detailPage.clickWorkItemDetailDescription()
         detailPage.setWorkItemDetailDescription (workItemUpdatedDescription, false);
@@ -126,7 +125,6 @@ var waitTime = 30000;
         //detailPage.clickWorkItemTitleEditIcon();
         detailPage.setWorkItemDetailTitle (workItemUpdatedTitle, false);
         detailPage.clickWorkItemTitleCancelIcon();
-
         detailPage.clickWorkItemDetailDescription()
         //detailPage.clickWorkItemDescriptionEditIcon();
         detailPage.setWorkItemDetailDescription (workItemUpdatedTitle, false);
@@ -143,8 +141,7 @@ var waitTime = 30000;
   });
 
 /* Verify that edits made to a workitem in the detail page, if cancelled, are discarded */
-
- it('should cancel edits to the workitem through its detail page - phone.', function() { 
+ it('should cancel edits to the workitem through its detail page - phone.', function() {
     /* Create a new workitem */
     var workItemTitle = "The test workitem title";
     var workItemUpdatedTitle = "The test workitem title - UPDATED";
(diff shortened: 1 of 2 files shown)


joshuawilson:
@naina-verma please rebase this

almightytest:
@joshuawilson ,
@sanbornsen , have helped me to do rebase , but now few tests are failing due to recent changes in "Description" input behavior. So i will update the PR with my commits

joshuawilson:
[test]

joshuawilson:
@ldimaggi please review

src/tests/work-item/work-item-list/displayAndUpdateWorkItemDetails.spec.js
```diff
@@ -140,32 +138,188 @@ var waitTime = 30000;
...
+
+        /* Fill in/update the new work item's title with blank and details field */
+        expect(page.workItemTitle(page.firstWorkItem)).toBe(workItemTitle);
+        page.workItemViewId(page.workItemByTitle(workItemTitle)).getText().then(function (text) {
+          var detailPage = page.clickWorkItemTitle(page.firstWorkItem, text);
+
+          detailPage.clickWorkItemTitleEditIcon();
+          detailPage.setWorkItemDetailTitle (workItemUpdatedTitle, false);
+          detailPage.clickWorkItemTitleSaveIcon();
+          expect((element(by.css(".clearfloat.alert.alert-danger"))).isPresent()).toBeTruthy();
```
ldimaggi: The current code is not incorrect, and it does run without error, but I think it needs a couple of changes.

First, we do not want any selector code in the tests. These should be all be placed in the page object model code so that the tests can be written at a higher level and will be easier to read and maintain. The "by.css" selector should be moved into the detail page object model code.

Second, while code will find the "alert-danger" UI element, it is too generic and could locate the wrong, or multiple UI elements. I'd suggest replacing this expect statement with 2 statements, each of which check for these new UI selectors: 

.//*[@id='wi-title-div'][.//*[contains(@class, 'alert-danger')]]
.//*[@id='wi-title-div']//p[.//text()[contains(.,'Title is required')]]

Yes - we'll need new code in the detail page object model for these too.   ;-)

src/tests/work-item/work-item-list/displayAndUpdateWorkItemDetails.spec.js
```diff
@@ -1,16 +1,16 @@
 /**
  * POC test for automated UI tests for ALMighty
- *  Story: Display and Update Work Item Details
```
ldimaggi: I think that your editor is adding spaces and newline chars.    ;-)

joshuawilson:
@ldimaggi please review
[test]

naina-verma:
@ldimaggi , updated Pr with your suggestions. please review

joshuawilson:
[test]

src/tests/work-item/work-item-list/work-item-detail.page.js
```diff
@@ -195,6 +195,11 @@ WorkItemDetailPage.prototype  = Object.create({}, {
     {
       return this.workItemDescriptionCancelIcon.click(); }
   },
+  titleValidation:   {
+    value: function ()
+    { 
+      return element(by.css(".clearfloat.alert.alert-danger")); }
```
ldimaggi: I think that we actually should have (2) separate checks for the alert. The current code is correct and works, but it does not verify the contents of the alert that is displayed. I'd suggest a 2-step approach:

```
  titleAlert:   {
    value: function ()
    { 
      return element(by.xpath(".//[@id='wi-title-div'][.//[contains(@class, 'alert-danger')]]")); }
  },


  titleAlertValidation:   {
    value: function ()
    { 
      return element(by.xpath(".//*[@id='wi-title-div']//p[.//text()[contains(.,'Title is required')]]")); }
  },

```

Other than this last suggested change - it looks great!

naina-verma:
@ldimaggi , Fixed ! please verify

joshuawilson:
[test]

PR #313, opened 2016-11-11
Attempt to install Patternfly SASS a bit better

pmuir:
DO NOT MERGE

An attempt to install patternfly properly. It's pretty obvious looking at the site and the number of hacks in the CSS that patternfly is not correctly working at the moment.

Previously, we were importing patternfly css directly, and then also importing it as SASS. The primary issue this caused was to mean the css priority ordering was screwed as we weren't loading the components of patternfly in the correct order, which caused some odd issues with styles getting randomly ignored - particularly as we were not importing bootstrap.

There was also an issue that we had set the font-path for patternfly etc. to point at node_modules, which meant that we were probably going to end up with odd problems - we want to make sure we are using the version that associates with the stylesheet.

I'm not really sure what I have done is correct at all, and there are still open issues:

1) With Angular 2 and SASS I still can't work out how we do a "global css" setup - we seem to have to import it in to every module, which seems odd to me. [1]

2) The Patternfly list view is screwed up. I suspect that because the styles were loading in the wrong order we ended up trying to fix this some other way.

3) We seem to have made various edits to patternfly itself which this commit removes (rightly) and I can't really figure out from the git history how to fix this - perhaps someone else knows?

4) I used Ruby to generate my patternfly install, and bundled the script - this is clearly A BAD IDEA in the long run as this project is not a ruby project, but hey, I know Ruby a lot better ;-)

5) I have no idea if this works in prod, as I didn't figure out yet how to run a prod build - I'm hoping this PR kicks off a build :-p

6) I seem to have slowed something down - probably related to (1)

Anyway, I hope this inspires someone to dig into a bit, as I'm somewhat at the end of my ability in fixing this!


joshuawilson:
This solution is on the right path but @pmuir found a better way and this PR is now obsolete.

PR #317, opened 2016-11-15
Making CLI awesome (plus perf metrics)

harrybabu:
A CLI dashboard for your webpack dev server, which lists out the assests and their size

[Webpack Dashboard](https://github.com/FormidableLabs/webpack-dashboard)

screen shot 2016-11-15 at 4 36 22 pm


joshuawilson:
I have been considering several performance tools. I think it would better to be optional instead of being on 'npm start', running every time I restart the app.
[test]

harrybabu:
@joshuawilson rebased and add new run command "start-dev" to display the dashboard, instead of default start.

Issue #311, opened 2016-11-11
Edit Icon Persistent on Title Field


mindreeper2420:
The Title field should have the same interaction as the description now does in production (11/11/16 @ 10:32am EDT). I.E. the icon appears on hover over the title, indicating that it is editable. The icon should not be persistent, as it takes up unnecessary space and is not consistent with the designs.

<img width="369" alt="screen shot 2016-11-11 at 10 33 43 am" src="https://cloud.githubusercontent.com/assets/4032718/20220131/58f29bf2-a7fa-11e6-8163-db51d506c72f.png">

nimishamukherjee:
Pr #318 merged hence closing this issue.

PR #318, opened 2016-11-15
Fix title edit icon persistence (#311)

nimishamukherjee:
Fix title edit icon persistence (#311)

### DIFF
--- a/src/tests/work-item/work-item-list/work-item-detail.page.js
+++ b/src/tests/work-item/work-item-list/work-item-detail.page.js
@@ -80,9 +80,28 @@ WorkItemDetailPage.prototype  = Object.create({}, {
   },
 
   /* Icon edit buttons */
-  workItemTitleEditIcon:  {
-    get: function ()
-    { return element(by.id("workItemTitle_btn_edit")); }
+  workItemTitleClick: {
+    value: function(){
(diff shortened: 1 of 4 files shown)


nimishamukherjee:
@ldimaggi could you have a look at the functional tests for this PR? I made some changes to them.
The UI behaviour change: 
On the details pane, the edit icon was visible next to the work item title. As per #311 we need to remove this edit icon and display it only on hover.

ldimaggi:
The test changes are fine - we will write a separate test to verify that the hovering is successful - thanks!

src/tests/work-item/work-item-list/work-item-detail.page.js
```diff
@@ -80,9 +80,35 @@ WorkItemDetailPage.prototype  = Object.create({}, {
   },
 
   /* Icon edit buttons */
-  workItemTitleEditIcon:  {
+  editHover:   {
     get: function ()
-    { return element(by.id("workItemTitle_btn_edit")); }
+    { 
+      return element(by.id(""));
```
joshuawilson: Does this ID need to reference something? (Or maybe I just don't understand how it works.)

joshuawilson:
@nimishamukherjee please rebase

Issue #237, opened 2016-10-25
404 on font resources on demo.almighty.io


aslakknutsen:
Currently, the following resources return 404:

http://demo.almighty.io/node_modules/patternfly/dist/fonts/PatternFlyIcons-webfont.ttf
http://demo.almighty.io/node_modules/patternfly/dist/fonts/PatternFlyIcons-webfont.woff

joshuawilson:
@SMahil can you look into this please.

ppitonak:
It seems that you need to be logged-in to reproduce.

PR #320, opened 2016-11-17
Fix 404 errors on font resources (#237)

nimishamukherjee:
Fix the 404 errors on font resources (#237).

Explanation:
The error is caused because of two variables in the file src/assets/stylesheets/variables.scss.
"fa-font-path" and "font-path" refer to node_modules/patternfly/dist/fonts and this causes the error as the node_modules folder does not exist on production.

In src/vendor.ts we are importing the patternfly csses.
import 'patternfly/dist/css/patternfly.min.css';
import 'patternfly/dist/css/patternfly-additions.min.css';

These csses have the Patternfly and Font Awesome details including the path in the PatternFly npm package context. The icon and font files under src/assets/stylesheets were over-writing this default behaviour. I have removed all references to "fa-font-path" and "font-path" variables.

Also, changing icon-font-name-fa to FontAwesome from fontawesome-webfont fixes the missing fonts on the header and navigation bar.

To verify this locally, after running npm start, rename the node_modules folder and refresh the browser.


joshuawilson:
While this may fix the issue on the surface, it is not a real solution. You are changing the core Patternfly SASS files.

nimishamukherjee:
@joshuawilson I haven't touched any of the PatternFly files. I have worked on the files which we had placed under src/assets/stylesheets.

joshuawilson:
Those files are downloaded from PF and should be untouched so they can be updated without breaking the app. Until we can use npm we manually have to update those files.

Issue #171, opened 2016-10-12
Kabob Menu Implementation


mindreeper2420:
According to the prototypes, the Kabob menu (the vertical ellipsis icon) controls the actions of the Work Item. These include moving the order of the item (to the top or bottom), Opening the item and deleting the item.
- Delete and View Details buttons should be removed from the list view.
- Opening an action from the Kabob that opens a modal (such as delete) should close the Kabob so that it is not still open.

InVision Prototype: https://redhat.invisionapp.com/share/YF8W6CC5Z#/screens/193140283

PatternFly: http://www.patternfly.org/pattern-library/widgets/#kebabs

sanbornsen:
Each action specified in the design is a task. I assume this issue is more about get the kebob menu in place and get it working properly. Am I right?

sanbornsen:
This issue is not one task, if I break it - 
- [ ]  Delete and View details button should be removed. (How to access the details page?)
- [ ]  Opening an action modal should close the kebob.
- [ ]  Move item top or bottom actions in the kebob (Need back-end support on this)

mindreeper2420:
@sanbornsen Accessing the Details page is done either by the 'Open' option in the kabob, or my clicking on the row itself (other areas reference showing the panel. There is not a separate details page).

joshuawilson:
@sanbornsen you can still implement moving items up and down with the kabob, you just can't persist the location. We do not have ordering setup yet.

sanbornsen:
@joshuawilson @mindreeper2420 I'd rather wait for the backend instead. Because anyway if we don't save the order, adding a new item will again re order it.

sanbornsen:
Last PR regarding this is merged - https://github.com/almighty/almighty-ui/pull/324
Hence closing the issue.

PR #324, opened 2016-11-18
Implement Move Top/Move Down options under the kabob menu, for a work item(#171).

nimishamukherjee:
Implement Move Top/Move Down options under the kabob menu, for a work item (#171). 
This is a UI implementation only as ordering has not been setup, the location cannot be persisted.
If we add a new item or refresh the page the ordering is lost.

### DIFF
--- a/src/app/work-item/work-item.service.ts
+++ b/src/app/work-item/work-item.service.ts
@@ -16,6 +16,7 @@ export class WorkItemService {
   private workItemTypeUrl = process.env.API_URL + 'workitemtypes';
   private availableStates: DropdownOption[] = [];
   public workItemTypes: WorkItemType[] = [];
+  private workItems: WorkItem[];
   
   constructor(private http: Http,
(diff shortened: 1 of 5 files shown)


joshuawilson:
@sanbornsen @harrybabu @SMahil please review

src/app/work-item/work-item.service.ts
```diff
@@ -60,6 +64,18 @@ export class WorkItemService {
       .catch(this.handleError);
   }
 
+  moveItem(wi: WorkItem, dir: String) {
```
nimishamukherjee: @sanbornsen  will add the comment

joshuawilson:
@nimishamukherjee the UX design uses the phrase "move to bottom" instead of "move down" and "move up" needs a "to" so that is it is "move to top"

nimishamukherjee:
@joshuawilson updated the texts.

PR #327, opened 2016-11-21
Double scroll handled (fixes #309)

### LINKED ISSUE #309: Two scrollbars are seen on the work item list page
Two scrollbars are seen on the work item list page - http://demo.almighty.io/work-item-list
twoscrollbars

sanbornsen:


src/app/app.component.html
```diff
@@ -1,3 +1,5 @@
+<div class="fix-container">
```
nimishamukherjee: The bottom of the scrollbar is not seen. If you try a resolution of 320x480, the scrollbar is not seen after add work item. 




Even in higher resolution, the bottom scroll arrow is not seen. Also the last work item is partially hidden.

sanbornsen: Yeah, so this would be another issue to fix the body.

sanbornsen: @SMahil you can take this up, create a new issue and work on it.
Also, we are not going to focus on mobile.

joshuawilson:
@SMahil please review and approve if you agree.

PR #334, opened 2016-11-22
Add Active State Identifier for WI Details

SMahil:


joshuawilson:
@SMahil is there an issue for this? Looking at the title I am unclear to what this fixes. We should have an issue for all PRs so that we can know what it fixes and track it in the open.

Issue #342, opened 2016-11-23
User assignee mobile view is broken


sanbornsen:
<img width="361" alt="screen shot 2016-11-23 at 9 43 34 am" src="https://cloud.githubusercontent.com/assets/2410474/20551040/c37592ac-b161-11e6-8916-d422e3db6cd9.png">

nimishamukherjee:
PR #343 merged. Issue resolved, hence closing the bug.

Issue #341, opened 2016-11-23
User assignee list design is broken if there is no name for the user


sanbornsen:
This is a very rare case scenario, but if there is no name for an user the design is broken.

<img width="270" alt="screen shot 2016-11-23 at 9 42 22 am" src="https://cloud.githubusercontent.com/assets/2410474/20551030/a1fa0928-b161-11e6-857e-b3a7eedb5685.png">

nimishamukherjee:
PR #343 merged. Issue resolved, hence closing the bug.

PR #343, opened 2016-11-23
User assignee css issues fixed

sanbornsen:
- (fixes-#341)
- (fixes-#342)

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
@@ -202,6 +202,7 @@
 }
 
 .user-search {
+  position: relative;
   .list-container{
     position: absolute;
     min-width: 200px;
@@ -233,6 +234,8 @@
       .item-li {
         cursor: pointer;
         padding: 5px;
+        @extend .width100;
+        float: left;
         &:hover {
           background-color: $color-pf-black-300;
         }
@@ -288,7 +291,7 @@
   .user-search {
     height: 100%;
     width: 100%;
-    position: relative;
+    position: fixed;
     z-index: 9;
     top: 0;
     left: 0;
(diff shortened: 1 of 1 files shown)


src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
```diff
@@ -288,14 +289,15 @@
   .user-search {
     height: 100%;
     width: 100%;
-    position: relative;
+    position: absolute;
```
joshuawilson: I don't think this is the right way to do this.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
```diff
@@ -233,6 +233,7 @@
       .item-li {
         cursor: pointer;
         padding: 5px;
+        height: 30px; // Because images are 20 + padding 5. It helps in no name scenario
```
SMahil: remove height and add width:100% and float:left;

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
```diff
@@ -288,14 +289,15 @@
...
-    position: relative;
+    position: absolute;
     z-index: 9;
     top: 0;
     left: 0;
     background-color: rgba(0, 0, 0, 0.6);
     .list-container {
       top: 12em;
       width: 80%;
+      position: relative;
```
SMahil: dont overwrite the position to relative and use transform to make it in center of div
we have mixin @include transform(translateX(-50%)); left:50%

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
SMahil: use position fixed as this behaviour acts as popup

PR #345, opened 2016-11-23
Funtional Tests for unauthorized user

naina-verma:
The major change in most of the files is due to change in structure of this workitem-list-page 
var WorkItemListPage = function (login) {  Added one argument for authorized and non-authorized user , its basically a Boolean value , which accept "true" for Authorized.
Test coverage:
-    Test if user is not authenticated can view almighty in read only
-     Test Quick add work item should not be visible
-     Test user should not be able update the work item type
-     Test user should not be able to update the work item state
-     Test user should not be able to Delete any work item

### DIFF
--- a/src/tests/work-item/work-item-list/unauthorizeduser.spec.js
+++ b/src/tests/work-item/work-item-list/unauthorizeduser.spec.js
@@ -1,0 +1,77 @@
+/**
+ * POC test for automated UI tests for ALMighty
+ *  Story: Display and Update Work Item Details
+ *  https://github.com/almighty/almighty-core/issues/296
+ *
+ * Note on screen resolutions - See: http://www.itunesextractor.com/iphone-ipad-resolution.html
+ * Tests will be run on these resolutions:
+ * - iPhone6s - 375x667
+ * - iPad air - 768x1024
+ * - Desktop -  1920x1080
+ *
+ * beforeEach will set the mode to phone. Any tests requiring a different resolution will must set explicitly.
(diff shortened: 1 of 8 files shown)


naina-verma:
@ldimaggi , can you please review!

joshuawilson:
[test]
@nimishamukherjee @sanbornsen please review

src/tests/work-item/work-item-list/unauthorizeduser.spec.js
```diff
@@ -0,0 +1,65 @@
...
+/*Test user should not be able update the work item title */
+  it('Test user should not be able update the work item title - phone.', function () {
+    testSupport.setBrowserMode('phone');
+    page.workItemByIndex(1).click();
+    expect(detailPage.workItemTitleEditIconById().isPresent()).toBeFalsy();
+   
+   });
+/*Test user should not be able update the work item description */
+  it('Test user should not be able update the work item description - phone.', function () {
+    testSupport.setBrowserMode('phone');
```
nimishamukherjee: Apart from the Edit Icon, the text is also clickable. Do you want to check that also?

naina-verma: @nimishamukherjee , can add this check in the same case! for title and description both.

naina-verma:
@ldimaggi , @nimishamukherjee  
Your requested changes has been done !
Please review

nimishamukherjee:
@joshuawilson this PR looks good to me.

joshuawilson:
[test]

PR #354, opened 2016-11-24
Make font awesome work on demo.almighty.io

nimishamukherjee:
Make font awesome work on demo.almighty.io
Renamed the icon-font-name-fa, in https://github.com/nimishamukherjee/almighty-ui/blob/alm-ui-353/src/assets/stylesheets/variables.scss, to FontAwesome.

This is in line with https://github.com/patternfly/patternfly/blob/master/src/less/variables.less


joshuawilson:
@nimishamukherjee how can I test this to confirm that it will fix an issue we only have in production?

joshuawilson:
"I don't test code often, but when I do, I test it in production"

PR #380, opened 2016-11-30
Help fix issues arising from storing of values by reference.

### LINKED ISSUE #370: The AssignTo List is some times missing 'me'
This is not directly a 'user is not signed in issue' since the navigation bar says 'logout' and the user is allowed to assign/edit the issue to begin with. Not sure how to reproduce it. (sporadically happens, possible related to some timeout?)

screenshot from 2016-11-29 14-45-10

### LINKED ISSUE #374: Bug: Missing User in Assign User Search
As a user logged into the system, I am unable to find myself by performing a search of the user list under the 'Assign User' feature in Work Item Detail.

screen shot 2016-11-29 at 8 48 43 pm

sanbornsen:
- Previously the data object from the service was stored in the
component variable.
- That used pass the reference of the object to the component variable.
- Changing anything on that in component was causing updating the data
object in the service.
- Now using lodash.cloneDeep() to copy the data in the component variable.
- fixes #370
- fixes #374

**npm run reinstall**


src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -89,8 +89,8 @@ export class WorkItemDetailComponent implements OnInit {
   getAllUsers() {
     this.userService.getAllUsers()
       .then((users) => {
-        this.users = users;
-        this.filteredUsers = users;
+        Object.assign(this.users, users);
```
joshuawilson: Did you know that this is only a shallow copy? It works for now but if the User has any children they will be references not copies.

sanbornsen: Ohh, thanks for identifying. I didn't realize that.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -89,8 +89,8 @@ export class WorkItemDetailComponent implements OnInit {
   getAllUsers() {
     this.userService.getAllUsers()
       .then((users) => {
-        this.users = users;
-        this.filteredUsers = users;
+        this.users = JSON.parse(JSON.stringify(users));
+        this.filteredUsers = JSON.parse(JSON.stringify(users));
```
joshuawilson: The problem with this approach is that it doesn't preserve the types. I think we may want cloneDeep<T>() from lodash.

sanbornsen: I agree, let's use lodash.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -2,6 +2,7 @@ import { Component, OnInit, Input, ViewChild, HostListener }      from '@angular
 import { ActivatedRoute, Params } from '@angular/router';
 import { Location }               from '@angular/common';
 import { Router }            from '@angular/router';
+import { cloneDeep } from 'lodash';
```
sanbornsen: @joshuawilson I am getting a warning in VS code on this line saying '/Users/sudiptasen/workspace/almighty-ui/node_modules/@types/lodash/index.d.ts' is not a module. But it works fine after compilation (npm start). Any idea? Am I missing anything in the config?

joshuawilson: try adding lodash to the 'types' in tsconfig.json

sanbornsen: That works, thanks.

sanbornsen:
Good to go !!

PR #395, opened 2016-12-01
Change detection strategy removed from header (fixes #371)

### LINKED ISSUE #371: Bug: Logged In User Icon Not Always Displayed
When navigating to demo.almighty.io on a browser that I have already logged into the site with, my user icon does not appear in the top right, as expected. It only appears when I navigate away from the Work Item List and then navigate back to it.

screen shot 2016-11-29 at 8 10 59 pm

sanbornsen:


src/app/header/header.component.ts
```diff
@@ -11,7 +11,6 @@ import { Broadcaster } from './../shared/broadcaster.service';
   selector: 'alm-app-header',
   templateUrl: './header.component.html',
   styleUrls: ['./header.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
```
sanbornsen: @joshuawilson any particular reason that you used change detection strategy on push?

PR #403, opened 2016-12-01
Prevent freezing if the target Workitem is not found

harrybabu:
Changes to prevent freezing of the screen if the target Workitem is not found while displaying the workitem links


joshuawilson:
[test]

joshuawilson:
While this fix works, it creates a new problem. Currently the UI does all the filtering, when we switch to the backend this fix will cause problems as it only displays items in the list.

PR #405, opened 2016-12-01
Prevent freezing if the target Workitem is not found

harrybabu:
Changes to prevent freezing of the screen if the target Workitem is not found while displaying the workitem links.

check if source/target is their in workitem list.


joshuawilson:
[test]

Issue #402, opened 2016-12-01
Match UI of Modal with Patternfly Design


SMahil:
Match UI of Modal/Dialog as per the Patternfly Design:

https://www.patternfly.org/pattern-library/widgets/#modal

joshuawilson:
This might be a great issue that can be broken out into smaller tasks. What do you think?

SMahil:
Yes we can break it down.

PR #413, opened 2016-12-02
FIXES: #402 =>Add transition and overlay for Dialog/Modal

SMahil:
For Dialog, Transition has been added and overlay has been set to the body. Earlier it was just showing or hiding the dialog box.

fixes  https://github.com/almighty/almighty-ui/issues/402

### DIFF
--- a/src/app/footer/footer.component.html
+++ b/src/app/footer/footer.component.html
@@ -1,1 +1,1 @@
-<div id="footer_container" class="footer width100 pull-left dn">Copyright 2016 Red Hat</div>
+<div id="footer_container" class="footer col-md-12 col-sm-12 col-xs-12 pull-left dn">Copyright 2016 Red Hat</div>
--- a/src/app/shared-component/dialog/dialog.component.ts
+++ b/src/app/shared-component/dialog/dialog.component.ts
@@ -1,9 +1,15 @@
 import {
+  animate,
   Component,
   EventEmitter,
   Input,
   OnInit,
-  Output
+  Output,
+  state,
+  style,
+  transition,
+  trigger
+  
 } from '@angular/core';
 
 import { Dialog } from './dialog';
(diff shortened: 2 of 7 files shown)


joshuawilson:
@SMahil this has failing tests. Please rebase and then maybe @naina-verma can help you.

SMahil:
@joshuawilson rebase done and tests are resolved

src/app/shared-component/dialog/dialog.component.ts
```diff
@@ -3,7 +3,7 @@ import {
   EventEmitter,
   Input,
   OnInit,
-  Output
+  Output, trigger, style , animate, state, transition
```
sanbornsen: Please write them in order and vertically.

src/app/shared-component/dialog/dialog.component.ts
```diff
@@ -12,6 +12,28 @@ import { Dialog } from './dialog';
...
+        transform: 'translateY(0)'
+      })),
+      transition('inactive <=>active', animate('500ms ease all'))
+    ]),
+    trigger('modalOverlay', [
+      state('inactive', style({
+        background: 'transparent'
+      })),
+      state('active', style({
+        background: 'rgba(0,0,0,0.4)'
```
SMahil: both have same output. If we are not following rgba we can change to hex with opacity

sanbornsen: @mindreeper2420 it's a background over lay that comes under the modal.

mindreeper2420: @SMahil just a consistency bit, IMO, but not a problem :)

@sanbornsen you can set those with opacity effects and layering, so makes no difference

SMahil: hex+alpha is not working over here

joshuawilson:
@SMahil please rebase

src/app/footer/footer.component.html
```diff
@@ -1 +1 @@
-<div id="footer_container" class="footer width100 pull-left dn">Copyright 2016 Red Hat</div>
+<div id="footer_container" class="footer col-md-12 col-sm-12 col-xs-12 pull-left dn">Copyright 2016 Red Hat</div>
```
nimishamukherjee: Do we need the footer, we always have quick add at the bottom.

SMahil: I just kept it rather than removing if we need it later sometime

joshuawilson:
@SMahil what issue/task is the suppose to fix? Please add the issue number to the Description at the top. Also please explain what this fix does?

PR #418, opened 2016-12-04
Add simple Comments to Work Item

aslakknutsen:
Create(secured) and List supported.

Attributes: body, created-at, created-by

almighty/almighty-core#348
almighty/almighty-core#349 

Depends on https://github.com/almighty/almighty-core/pull/546

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-comment/work-item-comment.component.html
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-comment/work-item-comment.component.html
@@ -1,0 +1,33 @@
+<div class="comment-container paddingT20" id="wi-comment" style="clear:both">
+    <div class="comment-content" id="wi-comment-content">
+        <!-- Refactoring needed to toogle view -->
+        <div *ngIf="true" class="width100">
+            <div class="width100  paddingL20 paddingR20" *ngIf="true">
+                <div *ngIf="loggedIn" class="paddingB20">
+                    <textarea rows="2" type="text" id="wi-comment-add-comment" class="form-control width100" [(ngModel)]="comment.attributes.body"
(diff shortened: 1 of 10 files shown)


src/app/work-item/work-item-list/work-item-detail/work-item-comment/work-item-comment.service.ts
```diff
@@ -0,0 +1,49 @@
...
+      this.headers.set('Authorization', 'Bearer ' + this.auth.getToken());
+    }
+    logger.log('WorkItemCommentService running in ' + process.env.ENV + ' mode.');
+  }
+
+  getComments(id: string): Promise<Comments> {
+    return this.http
+      .get(this.workItemUrl + '/' + id + '/relationships/comments', { headers: this.headers })
+      .toPromise()
+      .then((response) => process.env.ENV != 'inmemory' ? response.json() as Comments : response.json().data as Comments)
```
joshuawilson: do we need this if we are using the new api?

src/app/work-item/work-item-list/work-item-detail/work-item-comment/work-item-comment.component.html
```diff
@@ -0,0 +1,33 @@
+<div class="comment-container paddingT20" id="wi-comment" style="clear:both">
+    <div class="comment-content" id="wi-comment-content">
+        <!-- Refactoring needed to toogle view -->
+        <div *ngIf="true" class="width100">
+            <div class="width100  paddingL20 paddingR20" *ngIf="true">
+                <div *ngIf="loggedIn" class="paddingB20">
+                    <textarea #commentText rows="2" type="text" id="exampleDesc" class="form-control width100" [(ngModel)]="comment.attributes.body"
```
aslakknutsen: This is the 'create comment text area'. There will be only one, but yea, commentText is a bad name. Suggestions?

PR #424, opened 2016-12-05
Infinite scroll on work item list added

sanbornsen:
- Output events to update the list removed from quickadd.
- Output events to update the list removed from wi entry on delete.
- Using reference of the list object all over the places to update the dom.
- Added a new directive for handling infinite scroll event and page size.

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
@@ -213,7 +213,6 @@ export class WorkItemDetailComponent implements OnInit {
       .update(this.workItem)
       .then((workItem) => {
         this.workItem.version = workItem.version;
-        this.broadcaster.broadcast('updateWorkItem', workItem);
         this.activeOnList();          
     });
      
--- a/src/app/work-item/work-item-list/work-item-list.component.ts
+++ b/src/app/work-item/work-item-list/work-item-list.component.ts
@@ -1,5 +1,6 @@
 import { Component, OnInit, trigger, state, style, transition, animate } from '@angular/core';
 import { Router }            from '@angular/router';
+import { cloneDeep } from 'lodash';
 
 import { AuthenticationService } from './../../auth/authentication.service';
 import { Broadcaster } from './../../shared/broadcaster.service';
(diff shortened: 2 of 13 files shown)


src/app/work-item/work-item-list/work-item-list.component.ts
```diff
@@ -48,31 +50,41 @@ export class WorkItemListComponent implements OnInit {
...
   }
 
   // model handlers
 
-  reloadWorkItems(): void {
+  initWiItems(event: any): void {
+    this.loadWorkItems(event.pageSize);
+  }
+
+  loadWorkItems(pageSize: number = 20): void {
```
nimishamukherjee: pageSize is defined as 10 in infinitescroll.directive. Why is it reset to 20 here?

src/app/work-item/work-item.service.ts
```diff
@@ -28,18 +30,57 @@ export class WorkItemService {
...
+          }
+          this.workItems = response.json().data as WorkItem[];
+        }
         return this.workItems;
       })
       .catch(this.handleError);
   }
 
+  getMoreWorkItems(): Promise<any> {
+    console.log(this.nextLink);
```
nimishamukherjee: Remove the console.log statement

src/app/work-item/work-item.service.ts
```diff
@@ -83,11 +124,16 @@ export class WorkItemService {
...
+      .then(() => {
+        let deletedItemIndex = this.workItems.findIndex((item) => item.id == workItem.id);
+        // removing deleted item from the local list
+        this.workItems.splice(deletedItemIndex, 1);
+      })
       .catch(this.handleError);
   }
 
   create(workItem: WorkItem): Promise<WorkItem> {
+    console.log(workItem);
```
nimishamukherjee: Remove console.log statement

sanbornsen: Yeah, this one can go off.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -213,7 +213,6 @@ export class WorkItemDetailComponent implements OnInit {
       .update(this.workItem)
       .then((workItem) => {
         this.workItem.version = workItem.version;
-        this.broadcaster.broadcast('updateWorkItem', workItem);
```
sanbornsen: We are now updating values by reference not by events.

src/app/work-item/work-item-list/work-item-list.component.ts
sanbornsen: Also the pageSize 10 in the directive is just a default value.

PR #426, opened 2016-12-05
Default selection for dialog (fixes #258)

### LINKED ISSUE #258: Default selection on Confirm dialog component
When selecting 'delete' from the work item menu on the work item list a 'confirm dialog' pops up. 

One of the options should be 'default' selected and respond to keyboard input.

Either 'Cancel' is selected and you can do Arrow-Left + Enter to move to confirm and perform the action Or 'Confirm' is selected and you do Enter to confirm and perform the action.

And 'Esc' should be same as 'Cancel'

sanbornsen:
fixes #258

### DIFF
--- a/src/app/shared-component/dialog/dialog.component.ts
+++ b/src/app/shared-component/dialog/dialog.component.ts
@@ -2,6 +2,7 @@ import {
   animate,
   Component,
   EventEmitter,
+  HostListener,
   Input,
   OnInit,
   Output,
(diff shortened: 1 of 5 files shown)


joshuawilson:
@sanbornsen please rebase

src/app/shared-component/dialog/dialog.component.ts
```diff
@@ -39,4 +40,22 @@ export class DialogComponent implements OnInit {
     this.onClick.emit(val);
     this.modalFadeIn = false;
   }
+
+  @HostListener('window:keydown', ['$event'])
+  onActionKeyPress(event: any): void {
+    let currentDefaultIndex = this.dialog.actionButtons.findIndex(i => i.default);
```
joshuawilson: What is "default" here? I'm getting a Type error on it each time it is used in this method.
Error:(79, 74) TS2339:Property 'default' does not exist on type 'Object'.

sanbornsen: Default is a new property to the dialog action data. Please refer to this file src/app/shared-component/dialog/dialog.ts. You should not get the error.

joshuawilson: The reason for the error is that the class ActionItem is a private class and it's Type can't be seen by TypeScript. This makes TS think that "default" is of Type Object. To fix this, you just need to "export" the class.

PR #428, opened 2016-12-05
Fixed the bug of higlighting WI item from the list

### LINKED ISSUE #422: The 'select' area is larger than 'click to open detail' area
It's possible to click right at the beginning or at the end of the WorkItemList entry and achieve the 'Item' showing as 'selected', but not details being updated.

https://youtu.be/AMv7uMC8_ZM

sanbornsen:
- Clicking extreme edge of a work item list was selecting the item
but not actually opening it.
- fixes #422

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "ng2-bootstrap": "^1.1.16",
     "ng2-completer": "^0.2.3",
     "ng2-dragula": "^1.2.1",
-    "ng2-dropdown": "0.0.15",
     "patternfly": "3.8.1",
     "reflect-metadata": "^0.1.8",
     "rxjs": "5.0.0-beta.12",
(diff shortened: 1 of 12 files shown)


aslakknutsen:
I think this runs the same issue as https://github.com/almighty/almighty-ui/pull/423

The right side work item list entry menu stops working, as the click on the parent takes over.

sanbornsen:
@aslakknutsen Yes I just realized. Let me fix that. Thanks.

sanbornsen:
@nimishamukherjee I tried that, in that case you'll lose out the area around that three dots button. This seemed like a more elegant solution to be. But yeah, you could be right, let's discuss.

joshuawilson:
@nimishamukherjee @sanbornsen I had thought to ask @SMahil to look into adjusting the padding and seeing that would work. It feels like it might be harder but a better solution.

joshuawilson:
@sanbornsen what does adding ng2-dropdown do to solve this issue? Something like that should at least be explained in the description at the top of the PR.

sanbornsen:
@joshua anyway at some point we had to do this for better control over the dropdown directive.

joshuawilson:
@sanbornsen if it doesn't solve the issue then it doesn't belong here. You can make a new issue and do it there. It only confused the solution.

sanbornsen:
@joshuawilson it does solve the issue in a better way.

PR #461, opened 2016-12-07
moved toolbar padding to upper level

edewit:
re opened because of faulty rebase

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -14,3 +14,12 @@
 	border-top: 1px solid $color-pf-blue-300;
 	margin-top: 2%;
 }
+
+.app {
+	padding: em(96) 0 em(66) 0;
+	height: 100%;
+}
+
+@media (max-width: ($grid-float-breakpoint - 1)){
+	.app{ padding-top: em(132); }
+}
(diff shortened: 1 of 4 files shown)


edewit:
@SMahil I only moved the exiting padding from work item to upper level if there needs to be more padding in some cases this would have been an other issue IMO

joshuawilson:
Thanks @edewit 
[test]

src/app/app.component.scss
```diff
@@ -14,3 +14,12 @@
 	border-top: 1px solid $color-pf-blue-300;
 	margin-top: 2%;
 }
+
+.app {
+	padding: em(96) 0 em(66) 0;
```
SMahil: when the user is no logged in, padding-bottom should be 0. Currently we are using  .containerPadd-nologin{ padding-bottom: 0; }

As we have moved the padding to outer element , we need padding-bottom:0 in that scenario for same element

joshuawilson:
@SMahil can you take a look at this please. It appears that the position of contents of Dashboard and Learn has been pushed down. This is under the Home tab.

joshuawilson:
@edewit @SMahil I would rather move forward on this, so I merged it. Please submit new issues and PRs for things that need fixing.

PR #462, opened 2016-12-07
Obsidian integration

edewit:
re opened because of faulty rebase


edewit:
you can test by executing:
```bash
export FORGE_URL=http://generator-backend-obsidian.e8ca.engint.openshiftapps.com/forge
```

joshuawilson:
Thanks @edewit 

[test]

joshuawilson:
@edewit is that link to the backend something we can add to the code so we can deploy this PR to production?

joshuawilson:
@edewit I went ahead and merged this. The UI doesn't break hard, there is just a missing backend. So everything else still works. So I merged it.This will allow you to send updates easier as needed. 

Thanks for the contribution.

PR #484, opened 2016-12-09
Fixes #483 - Removes old Home tab image and links directly to the Dashboard

### LINKED ISSUE #483: Remove the old Home tab image of the Dashboard
Now that we have a real Dashboard we can remove the fake one.

joshuawilson:


joshuawilson:
[test]

PR #491, opened 2016-12-11
Supporintg muilti-assignee data model

sanbornsen:
- This commit is to support multi-assignee data model update of core.
- This commit does not have UI implementation of multi assignee.
- Updated the corresponding tests (both functional and unit) to support the updated model.

This PR is to support multi-assignee data model update in core PR - https://github.com/almighty/almighty-core/pull/571

### DIFF
--- a/src/app/models/work-item.ts
+++ b/src/app/models/work-item.ts
@@ -19,11 +19,11 @@ export class WorkItemAttributes {
 }
 
 export class WorkItemRelations {
-  assignee: {
+  assignees: {
     data: {
       id: string,
       type: string
-    }
+    }[]
   };
   baseType: {
     data: {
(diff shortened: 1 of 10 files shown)


aslakknutsen:
[test]

joshuawilson:
[test]

src/app/models/work-item.ts
```diff
@@ -19,11 +19,11 @@ export class WorkItemAttributes {
 }
 
 export class WorkItemRelations {
-  assignee: {
+  assignees: {
     data: {
       id: string,
       type: string
-    }
+    }[]
```
joshuawilson: Nicely done. Why doesn't comments need the array brackets?

sanbornsen: Ohh, yes that's a mistake. Comments should also be an array.

sanbornsen: Actually, no. Comments will have a data property which will be an array. I'll add it when I'll be working on comment resolver.

joshuawilson: I wasn't sure if it needed it, just asking.

PR #493, opened 2016-12-12
Filter user bug fixed

sanbornsen:
- Filter user was coming null instead of the id due to some change of order in the code.
- If logged out 'filer by assigned' to me should not work so added a lame logic there. We'll need proper UX specification.

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.ts
+++ b/src/app/work-item/work-item-list/work-item-list.component.ts
@@ -62,14 +62,9 @@ export class WorkItemListComponent implements OnInit, AfterViewInit {
   panelState: String = 'out';
   contentItemHeight: number = 65;
   pageSize: number = 20;
-  filters: any[] = [{
-    id:  1,
-    name: 'Assign to Me',
-    paramKey: 'filter[assginee]',
-    active: false,
-    value: null
-  }];
+  filters: any[] = [];
   allUsers: User[] = [] as User[];
+  authUser: any = null;
 
   constructor(
     private auth: AuthenticationService,
@@ -86,7 +81,6 @@ export class WorkItemListComponent implements OnInit, AfterViewInit {
   ngOnInit(): void {
     this.listenToEvents();
     this.loggedIn = this.auth.isLoggedIn();
-    this.setFilterValues();
     // console.log('ALL USER DATA', this.route.snapshot.data['allusers']);
     // console.log('AUTH USER DATA', this.route.snapshot.data['authuser']);
   }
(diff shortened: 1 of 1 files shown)


run_functional_tests.sh
```diff
@@ -13,14 +13,14 @@ node_modules/protractor/bin/webdriver-manager update
 echo -n Starting Webdriver and Selenium...
 (node_modules/protractor/bin/webdriver-manager start >>$LOGFILE 2>&1 &)
 # Wait for port 4444 to be listening connections
-while ! (ncat -w 1 127.0.0.1 4444 </dev/null >/dev/null 2>&1); do sleep 1; done
+while ! (nc -w 1 127.0.0.1 4444 </dev/null >/dev/null 2>&1); do sleep 1; done
```
kwk: Why do you have to rename this? On my computer `ncat` is a link to `nc`:

```
$ ll /usr/bin/nc /usr/bin/ncat
lrwxrwxrwx. 1 root root      4 19. Jul 11:15 /usr/bin/nc -> ncat
-rwxr-xr-x. 1 root root 396192 19. Jul 11:15 /usr/bin/ncat
```

run_functional_tests.sh
```diff
@@ -13,14 +13,14 @@ node_modules/protractor/bin/webdriver-manager update
...
 echo done.
 
 # Start the web app
 echo -n Starting Almighty development server...
 (node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --progress --host 0.0.0.0 --port 8088 >>$LOGFILE 2>&1 &)
 # Wait for port 8088 to be listening connections
-while ! (ncat -w 1 127.0.0.1 8088 </dev/null >/dev/null 2>&1); do sleep 1; done
+while ! (nc -w 1 127.0.0.1 8088 </dev/null >/dev/null 2>&1); do sleep 1; done
```
kwk: Same here, why rename?

run_functional_tests.sh
```diff
@@ -35,8 +35,8 @@ node_modules/protractor/bin/protractor protractor.config.js
 TEST_RESULT=$?
 
 # Cleanup webdriver-manager and web app processes
-fuser -k -n tcp 4444
-fuser -k -n tcp 8088
+kill -9 $(lsof -ti tcp:8088)
+kill -9 $(lsof -ti tcp:4444)
```
kwk: Why is this change needed? And why is the order changed? Previously we killed the process that owns port 8088 first.

joshuawilson:
@sanbornsen maybe you could have it say "Please Login" instead of being empty. It looks broken instead of providing direction.

run_functional_tests.sh
sanbornsen: ahh... my bad. That's my script for mac and I committed it. Thanks.

sanbornsen:
@joshuawilson yeah, I kept it that way because there could other filters that you can use without login. UX call.

joshuawilson:
This is to let @almighty/uxd know that this fix will remove "Assign to me" and leave it empty when the user is not logged in. At the moment this is the only filter and looks odd. However, when more filters are added I expect it to look fine. If you have suggestions on what to do regarding filters that require the user to be logged in, please create an issue for it.

PR #504, opened 2016-12-13
no more exists

baijum:
do not ignore it


sanbornsen:
Actually we might wanna keep it there, in case someone mess with their type script configuration and generate typings directory, and do a git add . then it'll pretty messed up.

joshuawilson:
@baijum I agree with @sanbornsen there is no value in removing it and it could cause harm if someone doesn't understand TypeScript and adds it.

PR #505, opened 2016-12-13
fixes: remove padding-bottom when user is logged out

SMahil:
when user is logged out, padding bottom should be zero to container

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.scss
+++ b/src/app/work-item/work-item-list/work-item-list.component.scss
@@ -7,41 +7,65 @@
     background-color: transparent;
   }
 }
-
-.containerPadd-nologin{ padding-bottom: 0; }
-
-.list-action {
-  height: em(50);
-  display: flex;
-  align-items: center;
-  background-color: $color-pf-black-300;
-  justify-content: space-between;
-  .list-type {
-    font-weight: 700;
-    margin-left: em(10);
+.detail-list-container{
+    padding-bottom: em(65);
+    height: 100%;
+    overflow: hidden;
+  .list-action {
+    height: em(50);
+    display: flex;
+    align-items: center;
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -1,15 +1,27 @@
...
+})
+export class AppComponent implements OnInit {
   
-  constructor(auth: AuthenticationService){
+  constructor(private auth: AuthenticationService, private broadcaster: Broadcaster){
     //auth.isLoggedIn();
   }
-}
\ No newline at end of file
+  loggedIn: Boolean = false;
```
joshuawilson: What are you trying to do here? How does this relate to fixing the padding for login?

Usually the app.component.ts file is for bootstrapping the entire application. So this kind of thing is not written here.

SMahil: [class.containerPadd-nologin] = "!loggedIn" I am toggling the state to know whether user is logged in or not. If not then removing the padding -bottom

sanbornsen: First of all, this padding was occurring in the list page, we should not add any logical customization over the main router.

SMahil: padding is being changed thats why this issue occurs again

joshuawilson:
@SMahil this needs to rebase too.

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -7,41 +7,65 @@
...
-  background-color: $color-pf-black-300;
-  justify-content: space-between;
-  .list-type {
-    font-weight: 700;
-    margin-left: em(10);
+.detail-list-container{
+    padding-bottom: em(65);
+    height: 100%;
+    overflow: hidden;
+  .list-action {
```
sanbornsen: Indentation looks incorrect here.

PR #506, opened 2016-12-13
Added Dashboard and learn page object model

naina-verma:
Added Dashboard and learn page object model

### DIFF
--- a/src/tests/work-item/work-item-list/page-objects/dashboard.page.js
+++ b/src/tests/work-item/work-item-list/page-objects/dashboard.page.js
@@ -1,0 +1,43 @@
+/**
+ * AlMighty page object example module for work item list page
+ * See: http://martinfowler.com/bliki/PageObject.html,
+ * https://www.thoughtworks.com/insights/blog/using-page-objects-overcome-protractors-shortcomings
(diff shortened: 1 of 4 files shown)


src/tests/work-item/work-item-list/page-objects/dashboard.page.js
```diff
@@ -0,0 +1,43 @@
...
+let CommonPage = require('./common.page');
+let constants = require("../constants");
+let until = protractor.ExpectedConditions;
+
+class DashboardPage {
+
+ constructor(login) {
+ };
+ welcomeWrapper () {
+    return element(by.css("welcomeWrapper"));
```
ldimaggi: Each of the UI elements that you define here should also have a "click" function - are they currently not click-able?

joshuawilson:
[test]

src/tests/work-item/work-item-list/page-objects/dashboard.page.js
ldimaggi: Ah - OK - let's just leave it at that for now. Thx!!

joshuawilson:
[test]

Issue #352, opened 2016-11-24
Dialog-based creation of work items


nimishamukherjee:
Develop the UI for dialog-based creation of work Items - https://github.com/almighty/almighty-core/issues/501

Mgranfie:
@nimishamukherjee noticed that the InVision link was not on this story. Let me know if this solution is something you have not seen yet... thanks.

https://redhat.invisionapp.com/share/Y39KFTAE7

nimishamukherjee:
PR #508 merged. Hence closing this issue.

PR #508, opened 2016-12-13
Add dialogue based creation of work items (#352)

nimishamukherjee:
Add dialogue based creation of work items (#352)

User story - https://github.com/almighty/almighty-core/issues/501
UX - https://redhat.invisionapp.com/share/Y39KFTAE7

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
@@ -17,7 +17,7 @@
             </li>
           </ul>
         </div>
-        <h5 class="detail-id pull-left" id="wi-detail-id">{{workItem.id}}</h5>
+        <h5 *ngIf="workItem.id" class="detail-id pull-left" id="wi-detail-id">{{workItem.id}}</h5>
         <span class="pficon-close detail-close" (click)="closeDetails()"></span>
       </div>
 
@@ -31,7 +31,7 @@
                 class="pficon-edit marginL10 fl detail-title-edit-ico" 
                 id="workItemTitle_btn_edit"></span>
           </div>
-          <div *ngIf="loggedIn && headerEditable" class="" id="wi-title-div"> 
+          <div *ngIf="(loggedIn && headerEditable || !workItem.id)" id="wi-title-div"> 
               <div almEditable 
                   (onUpdate)="checkTitle($event)"
                   (keyup.enter)="onUpdateTitle()"
@@ -66,7 +66,7 @@
 
         <div class="col-md-4 col-sm-5 col-xs-5 wi-state-btn pull-right padding0">
           <div dropdown class="dropdown card-pf-time-frame-filter detail-status-dropdown">
-            <button dropdown-open 
+            <button dropdown-open
               id="wi-detail-state"
               type="button" 
               class="btn dropdown-toggle col-md-12 col-sm-12 col-xs-12 btn-default">
@@ -90,20 +90,20 @@
       <div class="separator col-md-12 col-sm-12 col-xs-12 detail-assignee-wrap">
           <span class="pull-left user-label">Assignee</span>
           <span
-            *ngIf="!workItem.relationalData.assignees?.length"
+            *ngIf="!workItem.relationalData?.assignees?.length"
             (click)="activeSearchAssignee()" 
             class="pull-left fa fa-user-plus user-assign-icon">
           </span>
           <img 
-            *ngIf="workItem.relationalData.assignees?.length"
+            *ngIf="workItem.relationalData?.assignees?.length"
             (click)="activeSearchAssignee()"
             class="pull-left user-assign-avatar"
             src="{{workItem.relationalData.assignees[0]?.attributes.imageURL + '&s=20'}}" 
             onError="this.src='https://avatars0.githubusercontent.com/u/563119?v=3&s=20'" />          
           
           <div class="pull-left detail-assignee-name">
-              <span *ngIf="!searchAssignee && !workItem.relationalData.assignees?.length" (click)="activeSearchAssignee()" class="placeholder clickable"> Unassigned </span>   
-              <span *ngIf="!searchAssignee && workItem.relationalData.assignees?.length" (click)="activeSearchAssignee()" class="clickable" id="WI_details_assigned_user"> 
+              <span *ngIf="!searchAssignee && !workItem.relationalData?.assignees?.length" (click)="activeSearchAssignee()" class="placeholder clickable"> Unassigned </span>   
+              <span *ngIf="!searchAssignee && workItem.relationalData?.assignees?.length" (click)="activeSearchAssignee()" class="clickable" id="WI_details_assigned_user"> 
                 {{workItem.relationalData.assignees[0]?.attributes.fullName}} 
               </span>
               <div *ngIf="searchAssignee" class="user-search" >
@@ -149,24 +149,24 @@
           </div>
           <div class="pull-right col-md-6 col-sm-6 col-xs-12 no-padding">
             <span title="{{workItem.attributes['system.created_at'] | date:'fullDate'}}">
-              Created at: {{workItem.attributes['system.created_at'] | almMomentTime}}
-            </span>
+              Created at: {{ workItem.attributes['system.created_at'] ? (workItem.attributes['system.created_at'] | almMomentTime) : 'Creating now!'}}
+            </span>            
           </div>
       </div>
       <!--Creator-->      
       <div 
         class="width100 pull-left paddingL20 paddingR20 common-wrap">        
         <span class="pull-left user-label">Creator</span>
-          <span *ngIf="!workItem.relationalData.creator"
+          <span *ngIf="!workItem.relationalData?.creator"
              class="pull-left fa pficon-user user-creator-icon">
           </span>
-          <img *ngIf="workItem.relationalData.creator"            
+          <img *ngIf="workItem.relationalData?.creator"            
             class="pull-left user-avatar"
             src="{{workItem.relationalData.creator.attributes.imageURL + '&s=20'}}"
             id="WI_details_reporter_img" />          
           <div class="pull-left marginR10">                          
               <span id="WI_details_reporter_user"> 
-                {{workItem.relationalData.creator?.attributes.fullName || 'No creator found'}} 
+                {{workItem.relationalData?.creator?.attributes.fullName || 'No creator found'}} 
               </span>
           </div>        
       </div>
@@ -226,6 +226,8 @@
       </div>
     </fieldset>
   </form>
-  <alm-work-item-link [workItem]="workItem" [loggedIn]="loggedIn"></alm-work-item-link>
-  <alm-work-item-comment [workItem]="workItem" [loggedIn]="loggedIn"></alm-work-item-comment>
+  <div *ngIf="workItem.id">
+    <alm-work-item-link [workItem]="workItem" [loggedIn]="loggedIn"></alm-work-item-link>
+    <alm-work-item-comment [workItem]="workItem" [loggedIn]="loggedIn"></alm-work-item-comment>
+  </div>
 </div>
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
@@ -9,7 +9,7 @@ import { Broadcaster } from '../../../shared/broadcaster.service';
 import { Logger } from '../../../shared/logger.service';
 import { UserService } from '../../../user/user.service';
 
-import { WorkItem } from '../../../models/work-item';
+import { WorkItem, WorkItemAttributes, WorkItemRelations } from '../../../models/work-item';
 import { WorkItemService } from '../../work-item.service';
 import { WorkItemType } from '../../work-item-type';
 
@@ -52,6 +52,8 @@ export class WorkItemDetailComponent implements OnInit {
 
   loggedInUser: any;
 
+  addNewWI: Boolean = false;
+
   constructor(
     private auth: AuthenticationService,    
     private broadcaster: Broadcaster,    
@@ -72,22 +74,51 @@ export class WorkItemDetailComponent implements OnInit {
     this.route.params.forEach((params: Params) => {
       if (params['id'] !== undefined) {
         let id = params['id'];
-        this.workItemService.getWorkItemById(id)
-          .then(workItem => {
-            this.closeRestFields();
-            this.titleText = workItem.attributes['system.title'];
-            this.descText = workItem.attributes['system.description'];
-            this.workItem = workItem;
-            // fetch the list of user 
-            // after getting the Workitem
-            // to set assigned user 
-            // for this workitem from the list
-            this.getAllUsers();
-            this.activeOnList(400);
-          })
-          .catch (err => this.closeDetails());
+        if (id.indexOf('new') >= 0){
+          //Add a new work item          
+          this.addNewWI = true;
+          let type = id.substring((id.indexOf('?') + 1), id.length);
+          this.createWorkItemObj(type);
+          this.getAllUsers();          
+        }else{          
+          this.addNewWI = false;
+          this.workItemService.getWorkItemById(id)
+            .then(workItem => {
+              this.closeRestFields();
+              this.titleText = workItem.attributes['system.title'];
+              this.descText = workItem.attributes['system.description'];
+              this.workItem = workItem;
+              // fetch the list of user 
+              // after getting the Workitem
+              // to set assigned user 
+              // for this workitem from the list
+              this.getAllUsers();
+              this.activeOnList(400);
+            })
+            .catch (err => this.closeDetails());
+        }
       }
     });
+  }
+
+  createWorkItemObj(type: string) {
+    this.workItem = new WorkItem();
+    this.workItem.id = null;
+    this.workItem.attributes = new WorkItemAttributes();
+    this.workItem.relationships = new WorkItemRelations();
+    this.workItem.type = 'workitems';    
+    this.workItem.relationships = {
+      baseType: {
+        data: {
+          id: type,
+          type: 'workitemtypes'
+        }
+      }
+    } as WorkItemRelations;
+
+    this.workItem.attributes = {
+      'system.state': 'new'
+    } as WorkItemAttributes;
   }
 
   getAllUsers() {
@@ -204,14 +235,23 @@ export class WorkItemDetailComponent implements OnInit {
     }    
   }
 
-  save(): void {
-    this.workItemService
-      .update(this.workItem)
-      .then((workItem) => {
-        this.workItem.attributes['version'] = workItem.attributes['version'];
-        this.activeOnList();          
-    });
-     
+  save(): void {        
+    if (this.workItem.id){
+      this.workItemService
+        .update(this.workItem)
+        .then((workItem) => {
+          this.workItem.attributes['version'] = workItem.attributes['version'];
+          this.activeOnList();          
+      });
+    }else{
+      if (this.validTitle){
+        this.workItemService
+        .create(this.workItem)
+        .then((workItem) => {
+          this.router.navigate(['/work-item-list/detail/' + workItem.id]);
+        });
+      }      
+    }     
   }
 
   closeDetails(): void {
--- a/src/app/work-item/work-item-list/work-item-list.component.html
+++ b/src/app/work-item/work-item-list/work-item-list.component.html
@@ -1,5 +1,19 @@
-<div class="detail-list-container" [class.containerPadd-nologin] = "!loggedIn" >
-  
+<div *ngIf="loggedIn && showTypesOptions" class="types-modal">  
+  <h5>Click to select a type</h5>
+  <ul>
+    <li *ngFor="let type of workItemTypes" (click)="onChangeType(type.name)">      
+      <span almIcon [iconType]="type.name" class="xl-font"></span>
+      <div>{{type.name.substring(7, type.length)}}</div>              
+    </li>
+  </ul>
+  <span class="pficon-close detail-close" (click)="closePanel()"></span>
+</div>
+
+<div class="add-detailed-wi" (click)="showTypes()">
+  <span class="fa fa-plus icon"></span>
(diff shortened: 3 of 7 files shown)


src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -209,6 +212,8 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
       </div>
     </fieldset>
   </form>
-  <alm-work-item-link [workItem]="workItem" [loggedIn]="loggedIn"></alm-work-item-link>
-  <alm-work-item-comment [workItem]="workItem" [loggedIn]="loggedIn"></alm-work-item-comment>
-</div>
+  <div *ngIf="!addNewWI">
```
joshuawilson: Does this mean we can't add a link when we create a new workitem?

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -17,7 +17,7 @@
             </li>
           </ul>
         </div>
-        <h5 class="detail-id pull-left" id="wi-detail-id">{{workItem.id}}</h5>
+        <h5 *ngIf="!addNewWI" class="detail-id pull-left" id="wi-detail-id">{{workItem.id}}</h5>
```
sanbornsen: Can we just do it like if workItem.id available or not? So maybe replace {{workItem.id}} with
{{workItem?.id}} and remove *ngIf="!addNewWI".

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
nimishamukherjee: As per the wireframe, all the fields remain disabled until the title has been entered and saved. On the UI, the comments and link are not shown until the title is created.
Implementation wise, 
1. Create a work item with type + title
2. On save, we get the work item ID
3. We reload the details page with the newly created work item (this is the same as edit now)

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -66,7 +66,7 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
 
         <div class="col-md-4 col-sm-5 col-xs-5 wi-state-btn pull-right padding0">
           <div dropdown class="dropdown card-pf-time-frame-filter detail-status-dropdown">
-            <button dropdown-open 
+            <button dropdown-open [disabled]='addNewWI'
```
sanbornsen: Why do you need to keep it disabled until the item is added? I understand that changing it makes an API call for update but you can use the same logic in the event listener. In fact you can have all the fields like Description, States, Assignee, just don't let any of the vent listener call the save method until you have workItem.attributes['system.title']

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -89,19 +89,19 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
       </div>
       <div class="col-md-12 col-sm-12 col-xs-12 detail-assignee-wrap">
           <span
-            *ngIf="!workItem.relationalData.assignees?.length"
+            *ngIf="addNewWI || !workItem.relationalData.assignees?.length"
```
sanbornsen: Same logic as I explained up there. I don't think addNewWI flag is needed anywhere.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -89,19 +89,19 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
...
             (click)="activeSearchAssignee()" 
             class="pull-left fa fa-user-plus user-assign-icon">
           </span>
           <img 
-            *ngIf="workItem.relationalData.assignees?.length"
+            *ngIf="!addNewWI && workItem.relationalData.assignees?.length"
```
sanbornsen: Same logic as I explained up there. I don't think addNewWI flag is needed anywhere.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -89,19 +89,19 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
...
             (click)="activeSearchAssignee()"
             class="pull-left user-assign-avatar"
             src="{{workItem.relationalData.assignees[0]?.attributes.imageURL + '&s=20'}}" 
             onError="this.src='https://avatars0.githubusercontent.com/u/563119?v=3&s=20'" />
           
           
-          <div class="pull-left detail-assignee-name">
+          <div *ngIf="!addNewWI" class="pull-left detail-assignee-name">
```
sanbornsen: Same logic as I explained up there. I don't think addNewWI flag is needed anywhere.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -148,14 +148,17 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
               </div>
           </div>
           <div class="pull-right">
-            <span title="{{workItem.attributes['system.created_at'] | date:'fullDate'}}">
-              Created at: {{workItem.attributes['system.created_at'] | almMomentTime}}
+            <span *ngIf="!addNewWI" title="{{workItem.attributes['system.created_at'] | date:'fullDate'}}">
```
sanbornsen: Same logic as I explained up there. I don't think addNewWI flag is needed anywhere.
In this case you can just replace 'Oops do not have the time!' with 'Creating Now'

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -148,14 +148,17 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
...
+            </span>
+            <span *ngIf="addNewWI" title="{{workItem.attributes['system.created_at'] | date:'fullDate'}}">
+              Creating now
             </span>
           </div>
       </div>
 
       <div class="col-md-12 col-sm-12 col-xs-12 detail-description-wrap">
-        <div class="col-md-12 col-sm-12 col-xs-12 detail-description" *ngIf="loggedIn">
+        <div class="col-md-12 col-sm-12 col-xs-12 detail-description" *ngIf="loggedIn && !addNewWI">
```
sanbornsen: Same logic as I explained up there. I don't think addNewWI flag is needed anywhere.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -191,7 +194,7 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
             </div>
           </div>
         </div>
-        <div id="wi-detail-desc" class="col-md-12 description" *ngIf="!loggedIn">
+        <div id="wi-detail-desc" class="col-md-12 description" *ngIf="!loggedIn || addNewWI">
```
sanbornsen: Same logic as I explained up there. I don't think addNewWI flag is needed anywhere.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
sanbornsen: Same logic as I explained up there. I don't think addNewWI flag is needed anywhere.

sanbornsen: Well, this actually means we can't create a link or comment until the item is created. I guess this is fine.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -174,7 +177,7 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
                 #desc 
                 (onUpdate)="descUpdate($event)"
                 (keydown.enter)="onUpdateDescription()"
-                (keyup.enter)="preventDef($event)"
+                (keyup.enter)="preventDef($event)"
```
sanbornsen: Please remove the white space

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -52,6 +52,8 @@ export class WorkItemDetailComponent implements OnInit {
 
   loggedInUser: any;
 
+  addNewWI: Boolean = false;
```
sanbornsen: Probably you'll not need it.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -72,24 +74,53 @@ export class WorkItemDetailComponent implements OnInit {
...
-            this.activeOnList(400);
-          })
-          .catch (err => this.closeDetails());
+        if (id.indexOf('new') >= 0){
+          //Add a new work item          
+          this.addNewWI = true;
+          let type = id.substring((id.indexOf('?') + 1), id.length);
+          this.createWorkItemObj(type);
+          this.getAllUsers();
+          this.activeOnList(400);
```
sanbornsen: This line should not be here. This function requires a workItem ID which is not available at this point.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -72,24 +74,53 @@ export class WorkItemDetailComponent implements OnInit {
...
-            this.getAllUsers();
-            this.activeOnList(400);
-          })
-          .catch (err => this.closeDetails());
+        if (id.indexOf('new') >= 0){
+          //Add a new work item          
+          this.addNewWI = true;
+          let type = id.substring((id.indexOf('?') + 1), id.length);
+          this.createWorkItemObj(type);
+          this.getAllUsers();
```
sanbornsen: This function wouldn't be required following the code on template, but since I think choosing the assignee before saving the item should be implemented, this will be needed.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -205,12 +236,20 @@ export class WorkItemDetailComponent implements OnInit {
   }
 
   save(): void {
-    this.workItemService
-      .update(this.workItem)
+    if (this.addNewWI){
```
sanbornsen: This is only one place I think the flag is needed.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -205,12 +236,20 @@ export class WorkItemDetailComponent implements OnInit {
   }
 
   save(): void {
```
sanbornsen: I would recommend to separate save and create method for mainly two reason - 

- Better readability.
- In very near future we are going to support partial update, when this will become more complex.

actually let's discuss about it.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
sanbornsen: I just thought, even this can be avoided. 
Check for !this.workItem.id instead

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -72,24 +74,53 @@ export class WorkItemDetailComponent implements OnInit {
...
+              this.activeOnList(400);
+            })
+            .catch (err => this.closeDetails());
+        }
       }
     });
   }
 
+  createWorkItemObj(type: string) {
+    this.workItem = new WorkItem();
```
sanbornsen: Set this.workItem.id = null or false and use this as flag in places.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.ts
```diff
@@ -204,14 +235,23 @@ export class WorkItemDetailComponent implements OnInit {
...
+  save(): void {        
+    if (this.workItem.id){
+      this.workItemService
+        .update(this.workItem)
+        .then((workItem) => {
+          this.workItem.attributes['version'] = workItem.attributes['version'];
+          this.activeOnList();          
+      });
+    }else{
+      if (this.workItem.attributes['system.title']){
```
sanbornsen: Can you use **this.validTitle** flag here. So that if you write some special validation logic tomorrow in **isValid()** function, then that'll work here as well.

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,59 @@
 @import "../../../assets/stylesheets/color-variables";
 @import "../../../assets/stylesheets/custom";
 
+.add-detailed-wi{
```
sanbornsen: This file needs indentation.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
joshuawilson: I'm ok with that, just wanted to confirm.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
joshuawilson: @sanbornsen that only works for Add but we want each to update in Edit mode.

sanbornsen: @joshuawilson yeah, so in case of adding, no API should be fired until the title is valid. So, she updated the code and it looks fine now.

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,59 @@
...
+  text-align: center;
+  padding-top: em(8);
+  font-size: 20px;
+  position: absolute;
+  z-index: 10;
+  top: 98;
+  right: 150;
```
SMahil: top and right value has to be in em

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,59 @@
 @import "../../../assets/stylesheets/color-variables";
 @import "../../../assets/stylesheets/custom";
 
+.add-detailed-wi{
+  text-align: center;
+  padding-top: em(8);
+  font-size: 20px;
```
SMahil: font-size has to be in em. the function em(20); needs to be here

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,59 @@
...
+  position: absolute;
+  z-index: 10;
+  top: 98;
+  right: 150;
+  width: em(26);
+	height: em(26);
+	background: #0088ce;
```
vikram-raj: indentation missing.

src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -1,5 +1,24 @@
-<div class="detail-list-container" [class.containerPadd-nologin] = "!loggedIn" >
-  
+<div *ngIf="loggedIn && showTypesModal" class="types-modal">  
+  <h5>Click to select a type</h5>
+  <ul>
+    <li *ngFor="let type of workItemTypes" (click)="onChangeType(type.name)">
+      <a>
```
vikram-raj: Don't use a tag if no need of href attribute. Its behavior may change in different browser.

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,59 @@
...
+  color: #ffffff;
+  cursor: pointer;
+	-moz-border-radius: em(13);
+	-webkit-border-radius: em(13);
+	border-radius: em(13);  
+}
+.types-modal{
+  background-color: white;
+  position: absolute;
+  top: 95;
```
vikram-raj: top value unit missing.

src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -1,5 +1,22 @@
-<div class="detail-list-container" [class.containerPadd-nologin] = "!loggedIn" >
-  
+<div *ngIf="loggedIn && showTypesModal" class="types-modal">  
+  <h5>Click to select a type</h5>
+  <ul>
+    <li *ngFor="let type of workItemTypes" (click)="onChangeType(type.name)">      
+      <span almIcon [iconType]="type.name" class="xl-font"></span>
+      <br />
```
SMahil: Avoid <br/ >. We need to align it with width and float

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -44,7 +92,7 @@
   cursor: pointer;
 }
 
-@media (max-width: ($grid-float-breakpoint - 1)){
+@media (max-width: ($grid-float-breakpoint - 1)){  
   .containerPadd-nologin{ padding-bottom: 0; }
   .workItem_Overlay{
     background-color:$color-pf-black;
```
SMahil: overlay background has to translucent

SMahil:
@nimishamukherjee @joshuawilson Don't we need common dialog/modal layout across the project rather than creating modal again and again.

I have already used ng2-modal for other dialog. Even If we want to create our own dialog we need something like that

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,54 @@
...
+        height: em(180);
+        border: 1px solid #ededed;
+        text-align: center;
+        margin: em(20) em(10);
+        padding-top: em(50);
+        color: #000000;
+        cursor: pointer;
+      }
+    li:hover {
+      color:  #0088ce;
```
SMahil: color name has to be variable

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,54 @@
...
+  text-align: center;
+  padding-top: em(8);
+  font-size: em(20);
+  position: absolute;
+  z-index: 10;
+  top: em(59);
+  right: em(100);
+  width: em(26);
+	height: em(26);
+	background: #0088ce;
```
SMahil: color names has to be variables

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,54 @@
 @import "../../../assets/stylesheets/color-variables";
 @import "../../../assets/stylesheets/custom";
 
+.add-detailed-wi{
+  text-align: center;
+  padding-top: em(8);
```
SMahil: instead of padding-top we need to align it vertically and horizontally center. use this css for span

position:absolute; top:50%; left: 50%; @include(transform: translate(-50%,-50%));

src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -1,5 +1,22 @@
-<div class="detail-list-container" [class.containerPadd-nologin] = "!loggedIn" >
-  
+<div *ngIf="loggedIn && showTypesModal" class="types-modal">  
+  <h5>Click to select a type</h5>
+  <ul>
+    <li *ngFor="let type of workItemTypes" (click)="onChangeType(type.name)">      
+      <span almIcon [iconType]="type.name" class="xl-font"></span>
```
SMahil: both of these span needs to be in some div to align it in center

src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -1,5 +1,22 @@
-<div class="detail-list-container" [class.containerPadd-nologin] = "!loggedIn" >
-  
+<div *ngIf="loggedIn && showTypesModal" class="types-modal">
```
SMahil: I am not sure if this is a modal.

src/app/work-item/work-item-list/work-item-list.component.scss
nimishamukherjee: This change is not related to this PR. We need to raise another issue to fix this.

nimishamukherjee:
[test]

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,59 @@
...
+  ul{
+    list-style: none;    
+    li{      
+        float: left;
+        width: em(150);
+        height: em(180);
+        border: 1px solid $color-pf-black-200;
+        text-align: center;
+        margin: em(20) em(10);
+        padding-top: em(50);
```
SMahil: the content inside li should be aligned with transform rather than padding-top.

src/app/work-item/work-item-list/work-item-list.component.scss
```diff
@@ -1,6 +1,59 @@
...
+	-webkit-border-radius: em(13);
+	border-radius: em(13); 
+  span{
+    position:absolute; 
+    top:50%; 
+    left: 50%; 
+    @include transform (translate(-50%,-50%));
+  } 
+}
+.types-modal{
```
SMahil: is this a modal?? or a normal layover

nimishamukherjee: It is a normal layover.

src/app/work-item/work-item-list/work-item-list.component.scss
nimishamukherjee: When I try variations of transform: translate(-50%, -50%);  the li alignment goes wrong. I think the outer div and ul need to be set first. We can create a bug/enhancement to track this.

src/app/work-item/work-item-list/work-item-list.component.scss
sanbornsen: Line numbers 12, 13, 16, 17, 18 are not looking indented.

PR #509, opened 2016-12-13
Fixes typo in the `/intro` header

bartoszmajsak:


bartoszmajsak:
Another question - why all but one `routerLinkActive="active"` are place on `<li>`? Shouldn't it be in`<a>` like with the intro and the [docs](https://angular.io/docs/ts/latest/api/router/index/RouterLinkActive-directive.html)?

joshuawilson:
@bartoszmajsak that is a good question. I'm changing the navigation now, so i'll look into it.

PR #511, opened 2016-12-13
Some of the first tests written depend on the in-memory count of 14 w…

ldimaggi:
…ork items - these tests should be removed and should eventually be replaced with query tests for numbers of work items that match queries.


joshuawilson:
[test]

Issue #514, opened 2016-12-14
UI Breaks for Assignee Name and Create at TimeStamp


SMahil:
UI Breaks for Assignee Name and Create at TimeStamp in mobile View

<img width="264" alt="screen shot 2016-12-14 at 10 19 22 am" src="https://cloud.githubusercontent.com/assets/10396359/21169969/fc2510d8-c1e6-11e6-8cdb-39ff6dc45ff3.png">

PR #515, opened 2016-12-14
fixes:514 - Align User Assignee name and create at timestamp in mobile view

SMahil:
fixes: #514 - Align User Assignee name and create at timestamp in mobile view


joshuawilson:
[test]

PR #521, opened 2016-12-14
Work items cannot be identified in the browser unless the width of th…

ldimaggi:
…e browser is set to a minimum of 390 for chrome and 400 for firefox. We had previously been able to successfully set the screen width to 375.


joshuawilson:
[test]

PR #529, opened 2016-12-16
Work item linking re-worked

sanbornsen:
- Useful methods from link service is moved to work item service.
- Link service is removed.
- Old spect for work item linking unit test is removed.
- Since, most of the template doesn't rely on component variable, need to write new logics for unit testing.
- Work item link resolver added to the work item service.
- Work item link HTML component reworked.
- Filter on avaiable link type is added.


config/karma.config.js
```diff
@@ -43,6 +43,7 @@ module.exports = function (config) {
     colors: true,
     logLevel: config.LOG_INFO,
     autoWatch: false,
+    // browsers: ['Chrome', 'PhantomJS_custom'],
```
sanbornsen: Needs to be removed

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.html
```diff
@@ -1,25 +1,37 @@
...
+          <span id="wi-link-total">
+            {{workItem.relationalData.totalLinkCount}}
+          </span>
+          <span class="link-header-nav-text">
+            Total Links
+          </span>
+        </div>
+        <div class="link-header-nav" *ngFor="let linkType of workItem.relationalData.linkDicts">
+            <span>
+              {{linkType.count}}
```
naina-verma: @sanbornsen , Can you add id's for link types ,as you have added for total count. That would be great!

sanbornsen: @naina-verma sure thing.

nimishamukherjee:
1. Create a bug
2. Link this bug to a planner item which is not in the current view (you need to scroll a few pages to view this planner item)
3. The linked work item does not show up

src/app/pipes/alm-valid-link-types.pipe.ts
```diff
@@ -0,0 +1,25 @@
+import { Pipe, PipeTransform } from '@angular/core';
```
sanbornsen: We need the pipe to filter the work item link types based on current work item type. For example if current work item is a bug, it should have only 'blocks' as link type list.

sanbornsen:
@nimishamukherjee fixed the bug.

src/app/pipes/alm-valid-link-types.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+  transform(arr: any[], workItem: WorkItem): any {
+    let opLinkTypes: any[] = [];
+    arr.forEach((item) => {
+      if (item.relationships.source_type.data.id == workItem.relationships.baseType.data.id) {
+        opLinkTypes.push({
+          name: item.attributes['forward_name'],
+          linkId: item.id,
+          linkType: 'forward'
+        });
+      } else if (item.relationships.target_type.data.id == workItem.relationships.baseType.data.id){
```
joshuawilson: You have a check for the Source and for the Target if they match. What happens if neither match.

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
```diff
@@ -90,98 +85,54 @@ export class WorkItemLinkComponent implements OnInit, OnChanges {
         
     }
 
-    onSelectRelation(relation: LinkType): void{
+    onSelectRelation(relation: any): void{
```
joshuawilson: why is the 'any'?

sanbornsen: The data we are getting in this param is not an entire LinkType any more, it's some manipulated data like just the link type ID, name (forward or reverse name).

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
```diff
@@ -90,98 +85,54 @@ export class WorkItemLinkComponent implements OnInit, OnChanges {
...
-        }).catch((error) => {
-            console.log(error);
-        });
+        this.createLinkObject(
+           this.workItem['id'], 
+           this.selectedWorkItem['id'], 
+           this.selectedLinkType.linkId, 
+           this.selectedLinkType.linkType
+          );
+        const tempValue = {'data': this.link};
```
joshuawilson: Why is a 'temporary value' a 'constant'? Maybe this should be a 'let'?

src/app/pipes/alm-valid-link-types.pipe.ts
sanbornsen: Well, by the logic they has to match either one of them. If they are not then it's a bug in linking.

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
```diff
@@ -90,98 +85,54 @@ export class WorkItemLinkComponent implements OnInit, OnChanges {
...
-        .then((wItems) => {
-            let workItems = cloneDeep(wItems);
-            workItems.forEach((item: any) => {
-                this.workItemsMap[item.id] = item;
-            });
-            this.calculateTotal();
-        });
-    }
-
-    calculateTotal(): void{
```
joshuawilson: Why did you remove calculateTotal and updateCount?

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
```diff
@@ -90,98 +85,54 @@ export class WorkItemLinkComponent implements OnInit, OnChanges {
...
     }
 
-    onSelectRelation(relation: LinkType): void{
+    onSelectRelation(relation: any): void{
         this.selectedLinkType = relation;
-        this.searchWorkItem.searchType = relation.relationships.target_type.data.id;
+        // this.searchWorkItem.searchType = relation.relationships.target_type.data.id;
     }
 
     createLink(event: any = null): void {
```
sanbornsen: Yes, my bad. Over looked it.

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
sanbornsen: Because all these data manipulation is happening in the workItemService. 
Check models/work-item.ts file.
export class RelationalData {
  creator?: User;
  comments?: Comment[];
  assignees?: User[];
  linkDicts?: LinkDict[];
  totalLinkCount?: number;
}

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
sanbornsen: The naming is not correct here. It's parameters instead of temporary value. I'll update it.

src/app/pipes/alm-valid-link-types.pipe.ts
joshuawilson: so if you end up with a bug in the linking will this code blow up or fail gracefully?

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
joshuawilson: Can you use a type for that? Maybe a number or string? Just asking if it is possible.

src/app/pipes/alm-valid-link-types.pipe.ts
sanbornsen: If send a buggy link, it will not blow up, it'll just not show up the link in the list. If none of these conditions passes the link will not be pushed into the list.

src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.ts
sanbornsen: Actually it's an object and I don't have any type defined for that kind of object. I'll add one then. Let me see.

Issue #526, opened 2016-12-15
Empty drop down of Filter shows up, for unauthorized user


naina-verma:
Empty drop down for Filter shows up for unauthorized user
![screenshot from 2016-12-15 18-04-17](https://cloud.githubusercontent.com/assets/19172870/21224234/24f3fc58-c2f1-11e6-8702-808ffc5628a4.png)

nimishamukherjee:
PR #534 merged. Closing this bug.

PR #534, opened 2016-12-19
Display "No filters available" message for a user who has not logged in or has not filters available (#526).

nimishamukherjee:
Display "No filters available" message for a user who has not logged in or has not filters available (#526).

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.html
+++ b/src/app/work-item/work-item-list/work-item-list.component.html
@@ -22,7 +22,10 @@
         <span class="caret"></span>
       </button>
       <ul class="dropdown-menu dropdown-menu-right" role="menu" aria-labelledby="wi_filter_dropdown">
-        <li *ngFor="let filter of filters;" role="presentation" (click)="activeFilter(filter.id)">
+        <li *ngIf="!filters.length">          
+          <span>No filters available</span>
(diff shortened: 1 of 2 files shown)


src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -8,9 +8,13 @@ <h4 class="list-type">Backlog</h4>
         <span class="caret"></span>
       </button>
       <ul class="dropdown-menu dropdown-menu-right" role="menu" aria-labelledby="wi_filter_dropdown">
-        <li *ngFor="let filter of filters;" role="presentation" (click)="activeFilter(filter.id)">
+        <li *ngIf="filters.length == 0">          
+          <a>No filters available</a>
```
SMahil: this should come in span because <a> tag will come with its other stylings too

sanbornsen: I think the anchor tag is fine here. We want all the styles of it, don't we?

SMahil: Do we need clickable No filters?

nimishamukherjee: I used the <a> tag to retain the styling.

PR #535, opened 2016-12-19
Set height and width on list view assignee image (#525)

### LINKED ISSUE #525: No height,width wrapper defined for avatar in list view
No height,width wrapper defined for avatar in list view.
There should be container having defined height and width of avatar in wi list view.
screenshot from 2016-12-15 15-48-37
screenshot from 2016-12-15 15-51-24

nimishamukherjee:
Set height and width on list view assignee image to fix #525.

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list-entry/work-item-list-entry.component.scss
+++ b/src/app/work-item/work-item-list/work-item-list-entry/work-item-list-entry.component.scss
@@ -5,6 +5,14 @@
   cursor: pointer;
   font-size: rem(12);
   align-items: center;
+  .user-avatar{
+    height: em(20);
+    width: em(20);
+    img{
+      width: 100%;
+      height: 100%;
+    }
+  }
   .workItemList_workItemType {
     width: em(120);
     float: left;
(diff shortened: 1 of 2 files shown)


src/app/work-item/work-item-list/work-item-list-entry/work-item-list-entry.component.scss
```diff
@@ -5,6 +5,10 @@
   cursor: pointer;
   font-size: rem(12);
   align-items: center;
+  .user-avatar{
```
SMahil: this class should come for parent container with width and height. For img we need to give width:100% otherwise image will be cut

sanbornsen: Same opinion.

nimishamukherjee: Updated - moved user-avatar to the parent div and set the img height to 100%

nimishamukherjee: Added height: 100%;
Did not add overflow or scroll as the parent div has fixed width and height.

SMahil: overflow is not needed

PR #547, opened 2016-12-20
Added UI elements POM for comments tests

naina-verma:
Added UI elements POM for comments tests, which includes ID's for few elements with the index values.
These UI elements are generic :)


naina-verma:
@ldimaggi , can you please review

joshuawilson:
[test]

Issue #551, opened 2016-12-20
Remove extra space in Work Item Link List


SMahil:
there is more space for work item link list ul item

PR #552, opened 2016-12-20
fix: Remove extra padding from WI link List

SMahil:
Remove extra padding from WI link List

#551

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.html
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.html
@@ -99,8 +99,8 @@
             <div class="link-list-wrap" *ngIf="workItem.relationalData.linkDicts">
                 <h4 class="link-list-header">Links</h4>
                 <div *ngFor='let linkDict of workItem.relationalData.linkDicts'>
-                    <h5><strong>{{linkDict.linkName}}</strong></h5>
-                    <ul>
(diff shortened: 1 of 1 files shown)


src/app/work-item/work-item-list/work-item-detail/work-item-link/work-item-link.component.html
```diff
@@ -99,8 +99,8 @@ <h2 class="col-md-12 col-sm-12 col-xs-12 dib">Work Item Links</h2>
             <div class="link-list-wrap" *ngIf="workItem.relationalData.linkDicts">
                 <h4 class="link-list-header">Links</h4>
                 <div *ngFor='let linkDict of workItem.relationalData.linkDicts'>
-                    <h5><strong>{{linkDict.linkName}}</strong></h5>
-                    <ul>
+                    <h5 class="col-md-12 col-sm-12 col-xs-12 no-padding">{{linkDict.linkName}}</h5>
```
vikram-raj: We need some font weight for linkname. Because everything looks same now. 

Refer to screenshot.
localhost-8088-work-item-list-detail-18

joshuawilson: While we can set this, it is really up to the @almighty/uxd team to decide the font weight for each item.

bdellasc:
We'll get a story written for visuals for this piece of the UI.

Issue #571, opened 2016-12-22
In delete modal the Confirm button should have blue background on hover


SMahil:
in Delete confirmation dialog box, on hover confirm It should be in blue background

<img width="701" alt="screen shot 2016-12-22 at 1 58 39 pm" src="https://cloud.githubusercontent.com/assets/10396359/21419716/5726972c-c84f-11e6-8540-6454ab5c378a.png">

PR #573, opened 2016-12-22
fix: Add hover colour on confirm button in delete WI modal

SMahil:
fix: Add hover colour on confirm button in delete WI modal

#571


vikram-raj:
Just a thought: 
Can we add these type of css properties in custom.scss or in some other common scss file because in future we may need these same style at many places? So that we can reuse it.

SMahil:
We can add it later if all the buttons need same css...

vikram-raj:
Ok

PR #591, opened 2016-12-23
Validates WI title/desc datatypes and values, fixes #530

### LINKED ISSUE #530: "null" shows up while updating description workitem : "Firefox"
"null" shows up while updating description workitem : "Firefox"
screenshot attached 
screenshot from 2016-12-19 11-30-18
screenshot from 2016-12-19 11-30-29

debloper:
- Makes the template display blank text for falsy values of `description`
- Avoids creating WI with `null` type descriptions form quick-add
  - Updated the validator for both title and descriptions
  - Reassign description as "" if no value was provided (i.e. null)

Signed-off-by: Soumya Deb <debloper@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/591)

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
@@ -193,11 +193,10 @@
                 id="wi-detail-desc" 
                 class="description"
                 [innerText]="
-                  workItem.attributes['system.description'] ? 
+                  !!workItem.attributes['system.description'] ?
                     workItem.attributes['system.description'] : 
                     descEditable ? 
-                      workItem.attributes['system.description'] : 
-                      'Work item description.'">
+                      '' : 'Work item description.'">
               </p>
               <div class="edit-icon">
                 <span id="workItemDesc_btn_edit" class="pficon-edit" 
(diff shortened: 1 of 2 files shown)


src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -195,11 +195,10 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
                 id="wi-detail-desc" 
                 class="description"
                 [innerText]="
-                  workItem.attributes['system.description'] ? 
+                  !!workItem.attributes['system.description'] ?
```
sanbornsen: What is the benefit of adding two !s here?

debloper: It forces a boolean out of any `type` of variable. So no matter what value `var x` holds or whatever type is it, `!!x` is a boolean with similar truthy/falsiness. Also, it can trace nested objects well for undefined nested chains, instead of getting flat-tyres and throwing errors (say you wanna check for `abc.pnq.xyz` but `abc.pnq` isn't defined at that instance).

It avoids some conditional corner-cases and is generally a best practice to use it, as all you want is a boolean check, not a type+value comparison. To avoid similar corner cases without `!!` you can as well use `try...catch` but that's lengthier as well as less efficient way to handle it.

nimishamukherjee:
[test]

debloper:
@nimishamukherjee reported the behavior on #606 as this is the case with the title as well. Can we handle this particular issue for both the cases as part of another follow up patch?

We are going to need another patch for the title anyway, and I think it'll be best to identify and address these two cases of the same issue separately and not part of this patch. What do you think?

nimishamukherjee:
[test]

nimishamukherjee:
[test]

Issue #538, opened 2016-12-19
Cannot create a new workitem through the new detailed dialog


ldimaggi:
See referenced screencast - the dialog screen goes blank when the exit button is clicked.

https://dl.dropboxusercontent.com/u/12067552/Screencasts/ALMighty_new_wi_dialog.webm

nimishamukherjee:
@ldimaggi PR #596 merged. Please review.

naina-verma:
@nimishamukherjee , I have verified this issue on behalf of @ldimaggi .
It works !

nimishamukherjee:
Thanks @naina-verma ! Closing this issue.

PR #596, opened 2016-12-26
Allow editing of text on Firefox, when adding new work item via the dialog box (#538)

nimishamukherjee:
Allow editing of text on Firefox, when adding new work item via the dialog box (#538).
Added minimum height for the title box.

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
@@ -69,6 +69,7 @@
       font-size: em(22);
       outline: 0;
       line-height: 1.3;
+      min-height: em(17);
     }
     aside{
       padding: em(5) 0 0;


naina-verma:
@nimishamukherjee , i have checked it on demo.almighty.io , its still observed on firefox !

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.scss
```diff
@@ -69,6 +69,7 @@
       font-size: em(22);
       outline: 0;
       line-height: 1.3;
+      min-height: em(17);
```
sanbornsen: Does this work well for both chrome and ff?

nimishamukherjee: Yes I verified on both the browsers and then set the height. It matches the edit mode of existing work items.

Issue #602, opened 2016-12-27
"No creator found" shows up when fullName doesnot exists for github user


naina-verma:
"No creator found" shows up when fullName doesnot exists for github user
![screenshot from 2016-12-27 13-44-59](https://cloud.githubusercontent.com/assets/19172870/21494963/e6ab3ed6-cc3a-11e6-8244-7445d7f67e01.png)

PR #603, opened 2016-12-27
Creator not found occured when user FullName is null

naina-verma:
This PR is related to bug #602 .

---
This change is [Reviewable](https://reviewable.io/reviews/almighty/almighty-ui/603)

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
@@ -152,7 +152,7 @@
           </div>
       </div>
       <!--Creator-->      
-      <div 
+      <div *ngIf="workItem.relationalData?.creator"
         class="width100 pull-left paddingL20 paddingR20 common-wrap">        
         <span class="pull-left user-label" id="creator_label">Creator</span>
           <span *ngIf="!workItem.relationalData?.creator"
@@ -164,7 +164,7 @@
             id="WI_details_reporter_img" />          
           <div class="pull-left marginR10">                          
               <span id="WI_details_reporter_user"> 
-                {{workItem.relationalData?.creator?.attributes.fullName || 'No creator found'}} 
+                {{workItem.relationalData?.creator?.attributes.fullName}} 
               </span>
           </div>        
       </div>


src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -163,8 +163,11 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
             src="{{workItem.relationalData.creator.attributes.imageURL + '&s=20'}}"
             id="WI_details_reporter_img" />          
           <div class="pull-left marginR10">                          
-              <span id="WI_details_reporter_user"> 
-                {{workItem.relationalData?.creator?.attributes.fullName || 'No creator found'}}
```
nimishamukherjee: Checked on demo.almighty.io, the UI behaves correctly. As full name is null it displays "No Creator". Core needs to send the Github name if the user's full name is not available.

naina-verma: @nimishamukherjee , @aslakknutsen 
Just wanted to get more clarification on fullName behavior ,couple of questions below
- Is core setting up fullName's value to Gtihub name in case of ,FullName = null ?
- Or there is a field in DB for Github username which shows up instead of Fullname ,when its null?

debloper: I think the proper solution here is to allow **API to provide the username string as well**, and change line 167 into:

```js
{{ workItem.relationalData.creator.attributes.fullName || workItem.relationalData.creator.attributes.userName }}
```

I've removed the check for the existence of the nested objects, guessing that the returned successful API response for a WI will always have a creator - if I'm wrong to assume that, please add back the safety-guards (i.e. `workItem.relationalData?.creator?.attributes.keyName`)

naina-verma: @debloper , 
that looks good solution , but according to @nimishamukherjee , core is fixing this thing from backend , in this case if there is fullName = null they will return github username.
I am still waiting for @aslakknutsen for the backend implementation.
If am not wrong , you are suggesting workItem.relationalData.creator.attributes.userName to be replaced by "No creator found" ?

debloper: > If am not wrong , you are suggesting workItem.relationalData.creator.attributes.userName to be replaced by "No creator found" ?

Nope. I think "No creator found" is a terrible UX feedback here.

Also, I don't support filling up `fullName` with username, when `fullName` doesn't exist. Not only because that's counter-intuitive, but it's a borderline false information management.

naina-verma: So,
There could be another solution ,Not showing up "creator" when creator (does'nt exists ) WI is not yet created ? there is no point of showing up for every new WI creation for this !
Any input for this ? @debloper @nimishamukherjee

aslakknutsen: Backend as a temp solution will use GitHub Username as FullName if no FullName is set in GitHub. When we move to 'stronger' User Profile mgm than the GitHub data will be used as a default with an option for the User to change it on our side.

src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
```diff
@@ -164,7 +164,7 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
             id="WI_details_reporter_img" />          
           <div class="pull-left marginR10">                          
               <span id="WI_details_reporter_user"> 
-                {{workItem.relationalData?.creator?.attributes.fullName || 'No creator found'}} 
+                {{workItem.relationalData?.creator?.attributes.fullName}} 
               </span>
```
nimishamukherjee: What do you do when the creator's data is not available?

naina-verma: @nimishamukherjee , i want to make this PR on hold #605 , waiting for exact UX suggestion.
Please ignore recent changes.
Thanks

Issue #558, opened 2016-12-21
Move work items top and bottom for order of execution 


nimishamukherjee:
Move work items top and bottom for order of execution 

Story - almighty/almighty-core#318
UX - https://redhat.invisionapp.com/share/5D9JPDJXE#/screens/209117619

nimishamukherjee:
PR #609 merged hence closing this task.

PR #609, opened 2017-01-02
Move work items top and bottom for order of execution  (#558)

nimishamukherjee:
This PR will add the UI behaviour only, for moving a work item to either the top or bottom of the visible screen.
UI Task - #558
Note: Current implementation of Move to Bottom will take the user to the end of the loaded list.

### DIFF
--- a/src/app/work-item/work-item.service.ts
+++ b/src/app/work-item/work-item.service.ts
@@ -435,16 +435,24 @@ export class WorkItemService {
    * ToDo: Integrate backend when available, also move by one
    * place should be implemented
    */
-  moveItem(wi: WorkItem, dir: String): void {
+  moveItem(wi: WorkItem, dir: String): Promise<any> {
     let index = this.workItems.findIndex(x => x.id == wi.id);
     wi.attributes.nextitem = '';
     wi.attributes.previousitem = '';
     switch (dir){
       case 'top':
-        //this.workItems.splice(0, 0, wi);
+        //move the item as the first item
+        this.workItems.splice(0, 0, wi);
+        //remove the duplicate element
+        this.workItems.splice( index + 1, 1);
+        wi.attributes.nextitem = parseInt(this.workItems[1].id);
         break;
       case 'bottom':
-        //this.workItems.splice(this.workItems.length, 0, wi);
+        //move the item as the last of the loaded list
+        this.workItems.splice((this.workItems.length), 0, wi);
+        //remove the duplicate element
+        this.workItems.splice( index, 1);
+        wi.attributes.previousitem = parseInt(this.workItems[this.workItems.length-2].id);
         break;
       case 'up':
         if (index > 0) { //no moving of element if it is the first element
@@ -480,6 +488,7 @@ export class WorkItemService {
     //console.log(wi.attributes.previousitem, ':' , wi.attributes.nextitem);
     //build the map to reset the indices
     this.buildWorkItemIdIndexMap();
+    return Promise.resolve();
     /*
     return this.http
       .patch(this.reorderUrl, JSON.stringify({data: wi}), { headers: this.headers })
--- a/src/app/work-item/work-item-list/work-item-list.component.html
+++ b/src/app/work-item/work-item-list/work-item-list.component.html
@@ -28,10 +28,10 @@
       </button>
       <ul class="dropdown-menu dropdown-menu-right" role="menu" aria-labelledby="wi_filter_dropdown">
         <li>
-          <a role="menuitem" tabindex="-1">Move to Top</a>
+          <a role="menuitem" tabindex="-1" (click)='onMoveToTop(workItemToMove)' > Move to Top </a>
         </li>
         <li>
-          <a role="menuitem" tabindex="-1">Move to Bottom</a>
+          <a role="menuitem" tabindex="-1" (click)='onMoveToBottom(workItemToMove)' > Move to Bottom </a>
         </li>
         <li role="presentation" class="divider"></li>
         <li>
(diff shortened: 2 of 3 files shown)


joshuawilson:
@nimishamukherjee will this fix also focus the list on the moved item? @almighty/uxd can you confirm the desired behavior, please?

nimishamukherjee:
@joshuawilson / @almighty/uxd , current behaviour:
when we click on the kebab menu, the work item gets selected and remains selected after being moved to the top or bottom.
1. Select a work item to move
move1

2. Move the work item to top
move2

3. Move the work item to the bottom
move3

joshuawilson:
@almighty/uxd can you confirm that 'move to top' and 'move to bottom' is only moving to top and bottom of the viewable area?

mindreeper2420:
@nimishamukherjee @almighty/uxd If the row is being displayed as selected, then the Detail View should be shown. After the row has been moved, it should fade back to de-selected.

joshuawilson:
@mindreeper2420 what about my question of "what is the bottom"? Bottom of the viewport or bottom of the list?

mindreeper2420:
@joshuawilson the bottom is the bottom of the loaded list.

joshuawilson:
@mindreeper2420 ok, if we have a total list of 100 and we are at the top of the list, we have loaded 20 items. If I tell the item at position 5 to move to the bottom, does it go to 20 or 100? What you said is the loaded list and that is 20. I would expect it to go to 100. 

I have a similar expectation that if I'm on 72 and I tell it to move to the top it should 1.

nimishamukherjee:
@mindreeper2420 - "If the row is being displayed as selected, then the Detail View should be shown. After the row has been moved, it should fade back to de-selected."

Few questions around this:
1. Does selecting the check-box imply selection of work item + opening of detailed view? Would the reverse hold good - deselecting the check-box would deselect the work item and close the detailed view?

2. Does selecting a work item for viewing the details select the check-box and display the top move button?

3. If the user wants to move a work item up by 3 rows, he would have to select the work item and click on "move up" three times. In this scenario, how would the detail view behave, if it was opened initially?

mindreeper2420:
@nimishamukherjee 

1. Correct - selecting the checkbox does not open the detail view.

2. Correct - the move dropdown is not enabled, as nothing has been selected to move. The move functionality in the kabob is neither enabled or disabled, though, so that is just there, but hidden behind the detail panel.

3. If the detail view is open at the time of selection and movement, then it would remain open until closed by the user. Moving items in the list does not force the detail panel to close.

nimishamukherjee:
@mindreeper2420 need your input on @joshuawilson 'squestion:

> @mindreeper2420 ok, if we have a total list of 100 and we are at the top of the list, we have loaded 20 items. If I tell the item at position 5 to move to the bottom, does it go to 20 or 100? What you said is the loaded list and that is 20. I would expect it to go to 100. I have a similar expectation that if I'm on 72 and I tell it to move to the top it should 1.

mindreeper2420:
@nimishamukherjee @joshuawilson a question I have before this is answered: is there a mechanism that informs the user there are more items that what have been loaded? In a standard paginated list, that would be easy with the pagination controls in PF. With the single-page list, there is normally a 20 of 100, etc. Is the UI showing that information?

If yes, then sending the item to the bottom makes sense to go the number 100. If not, then sending it to 100 when only 20 are visible (and perhaps known) would make it confusing and frustrating.

joshuawilson:
@mindreeper2420 there is support for display "20 of 100 items loaded", we just didn't have a requirement nor a design for it yet. I think we should add the "of total" and make move to bottom be the actual end of the list.

mindreeper2420:
@joshuawilson agreed - I think that makes the most sense.

Mgranfie:
This was an an early design and when the implementation went to pagination
and then back to infinite scroll, it go lost in the shuffle.

The number is located here in the Visual Designs.

https://redhat.invisionapp.com/share/NU8YPHLJP#/195933683_A-0100

It needs to be tweaked to fall into place with our current layouts. It
would be great if Brian and Nimishia could connect on this.

Thanks All
-Monica

nimishamukherjee:
[test]

vikram-raj:
@nimishamukherjee When I move WI to top or bottom and then click on that WI then It's open wrong detail page. 

It is working fine for the first time.

nimishamukherjee:
Thanks @vikram-raj good catch

src/app/work-item/work-item.service.ts
```diff
@@ -435,16 +435,24 @@ export class WorkItemService {
    * ToDo: Integrate backend when available, also move by one
    * place should be implemented
    */
-  moveItem(wi: WorkItem, dir: String): void {
+moveItem(wi: WorkItem, dir: String): Promise<WorkItem[]> {
```
sanbornsen: Indentation missing.

sanbornsen: Do you need to return WorkItem[] as promise? Can't it just be void?

src/app/work-item/work-item.service.ts
```diff
@@ -480,6 +488,9 @@ export class WorkItemService {
     //console.log(wi.attributes.previousitem, ':' , wi.attributes.nextitem);
     //build the map to reset the indices
     this.buildWorkItemIdIndexMap();
+    return new Promise((resolve, reject) => {
```
sanbornsen: Is this necessary?

src/app/work-item/work-item.service.ts
```diff
@@ -504,7 +513,7 @@ export class WorkItemService {
         return updatedWorkItem;
       })
       .catch (this.handleError);*/
-  }
+    }
```
vikram-raj: Extra space.

PR #615, opened 2017-01-03
Username pipe implemented

### LINKED ISSUE #602: "No creator found" shows up when fullName doesnot exists for github user

sanbornsen:
- This pipe will handle the display of user name all over the templates
- fixes #602

### DIFF
--- a/src/app/pipes/alm-user-name.pipe.ts
+++ b/src/app/pipes/alm-user-name.pipe.ts
@@ -1,0 +1,20 @@
+import { User } from './../models/user';
+import { Pipe, PipeTransform } from '@angular/core';
+
+// tslint:disable-next-line:use-pipe-transform-interface
+@Pipe({ name: 'almUserName', pure: true })
+export class AlmUserName implements PipeTransform {
+  transform(userObj: User | null = null, notFound: string = 'User not found'): string {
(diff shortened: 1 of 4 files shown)


src/app/pipes/alm-user-name.pipe.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Pipe, PipeTransform } from '@angular/core';
+
+// tslint:disable-next-line:use-pipe-transform-interface
+@Pipe({ name: 'almUserName', pure: true })
+export class AlmUserName implements PipeTransform {
+  transform(userObj: User | null = null, notFound: string = 'User not found'): string {
+    if (typeof(userObj) === 'undefined' || userObj === null) {
+      return notFound;
+    }
+    return userObj.attributes.fullName;
```
nimishamukherjee: What if userObj exists but fullName is null or undefined or ''?

sanbornsen: If user object is there then fullName has to be there. If it's null then the output should be empty string. I haven't handle that though. Updating.

PR #618, opened 2017-01-03
Tests for creator detail view page

naina-verma:
Tests covered : 
- Creator read only test
- On creating a work item creator should reflect 
- Creator image should be visble on creating work item (verification step)
Added user0 in the mock data , because of WI to reflect some user .

### DIFF
--- a/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
+++ b/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
@@ -298,6 +298,48 @@ class WorkItemListPage {
 
 
   }  
+  userStoryIconWIT  (){
+    return element(by.css('.xl-font.fa.fa-bookmark'));
+  }
+  clickUserStoryWItype  ()  {
+    return this.userStoryIconWIT.click();
+  }
+  valuePropositonWItype  ()  {
+   return element(by.css('.xl-font.fa.fa-gift'));
+  }
+  clickValuePropositonWItype  ()  {
+    return this.valuePropositonWItype.click();
+  }
+  fundamentalWItype  ()  {
+   return element(by.css('.xl-font.fa.fa-bank'));
+  }
+  clickFundamentalWItype  ()  {
+    return this.fundamentalWItype.click();
+  }
+  plannerWItype(){
+    return element(by.css('.xl-font.fa.fa-paint-brush'));
+  }
+  clickPalnnerWItype  ()  {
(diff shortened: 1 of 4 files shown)


src/tests/work-item/work-item-list/work-item-detailed-dialog.spec.js
```diff
@@ -27,10 +27,61 @@ describe('Work item list', function () {
...
+      expect(detailPage.getCreatorAvatar().isPresent()).toBe(true);  
+      expect(detailPage.getImageURL()).toBe('https://avatars.githubusercontent.com/u/2410471?v=3&s=20');
+   });
+ });
+ it('check Creator is readonly - desktop', function () {
+   page.clickDetailedDialogButton();
+   var detailPage = page.clickDetailedIcon("userstory");
+   detailPage.clickCreatorDefaultIcon();
+   expect(detailPage.getCreatorDefaultIcon().isPresent()).toBe(true);
+   expect(detailPage.getCreatorUsername()).toBe('No creator found');
```
ldimaggi: It looks like the error generated by the UI changed - the test will have to also change to stayin synch:

 ```
Work item list check Creator is readonly - desktop
  Message:
    Expected 'Creator not found' to be 'No creator found'.
```

File: src/tests/work-item/work-item-list/page-objects/work-item-list.page.js [hunk 2]
```diff
@@ -298,6 +298,27 @@ class WorkItemListPage {
 
 
   }  
+  clickUserStoryWItype  ()  {
```
ldimaggi: These functions are all syntactically correct and will run cleanly, but it is a better practice to define (2) functions for each UI element. 

For example:

The first function will define the UI element:

```
valuePropositonWItype  ()  {
   return element(by.css('.xl-font.fa.fa-gift'));
}
```

And the second function will invoke that UI element:

```
clickValuePropositonWItype  ()  {
   return this.valuePropositonWItype.click();
}
```

naina-verma: Hi @ldimaggi ,
I used this approach to save lengthy codes and re-usability  of one generic method. But if above approach looks good . I have modified the methods with your suggested example .
Thanks :)

naina-verma:
@ldimaggi , done with the changes !

nimishamukherjee:
[test]

Issue #604, opened 2016-12-28
Tittle display height should be little increase , words like "y,g" shows up incomplete 


naina-verma:
Tittle display height should be little increase , words like "y,g" shows up incomplete 
![screenshot from 2016-12-28 13-10-27](https://cloud.githubusercontent.com/assets/19172870/21516540/320e347c-ccff-11e6-894d-3bba30f1488c.png)

nimishamukherjee:
@naina-verma can you verify? PR #625 has been merged.

naina-verma:
@nimishamukherjee  Verified !

PR #625, opened 2017-01-04
Bug: Fix line height for WI Title in detail page

SMahil:
Bug: Fix line height for WI Title in detail page
#604


debloper:
@almighty/uxd can we have a styleguide of some sort?

I feel like most of the styling we have for now is somewhat random/magic numbers with eye-estimation that works for the time being, and to regress horribly, should the peer DOM styles may change.

mindreeper2420:
@debloper please review PatternFly.org for styles, etc.

Pattern Library: http://www.patternfly.org/pattern-library/#_
Styles: http://www.patternfly.org/styles/#_

If there is something that does not fall within the guidelines of PatternFly, then that needs to be brought up with the CSS Devs.

PR #627, opened 2017-01-04
Uses the new WI flag to prevent 2xTitle while saving, fixes #594

### LINKED ISSUE #594: Two titles are displayed during the dialog based work item creation
- Click on +
- Select any type
- Enter work item title
- Hit enter or click on tick to save
- For a few seconds we see two titles

debloper:
Find out the discussions in #594

Signed-off-by: Soumya Deb <debloper@gmail.com>

### DIFF
--- a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
+++ b/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html
@@ -34,7 +34,7 @@
                 class="pficon-edit marginL10 fl detail-title-edit-ico" 
                 id="workItemTitle_btn_edit"></span>
           </div>
-          <div *ngIf="(loggedIn && headerEditable || !workItem.id)" id="wi-title-div"> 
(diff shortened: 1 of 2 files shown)


joshuawilson:
[test]
Please rebase this

nimishamukherjee:
@debloper the issue still persists - the problem lies in https://github.com/debloper/almighty-ui/blob/0afc2dd5cf35646df969565d296ca6267d53601a/src/app/work-item/work-item-list/work-item-detail/work-item-detail.component.html#L32

Add some dummy text before {{ workItem.attributes['system.title'] }} and test the behaviour when 
1. Adding a work item using the dialogue box
2. Editing a work item

debloper:
Updated the patch; solving the issue another way.
Instead of making the `ngIf` checks more complicated (they're kinda fine, if process flow is counted out), I've forced a flag reset (`headerEditable`) while creating new WI which will work with the existing checks just fine.

nimishamukherjee:
[test]

PR #640, opened 2017-01-06
Planner as an NPM library package

michaelkleinhenz:
Planner as an NPM library package

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 /app/*.js
 /app/*.js.map
 /tmp
+/dist-library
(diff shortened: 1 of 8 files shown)


nimishamukherjee:
Screenshot of the error
npm-error

The UI works fine despite these error.

michaelkleinhenz:
The error message is a know problem with the existing build system. It does not interfere with the runtime, the compiler just "sees" the files that should be outside of it's scope. This will be fixed with the build system updates in the next sprint.

PR #641, opened 2017-01-06
Fixed link mocks.

michaelkleinhenz:
Fixes a problem with the mock of links. Extends the functionality following the core api to retrieve single sets of links connected to a work item.

### DIFF
--- a/src/app/shared/mock-data.service.ts
+++ b/src/app/shared/mock-data.service.ts
@@ -157,7 +157,7 @@ export class MockDataService {
             ]
           };
         }
-      }
+      } 
       // should never happen
       return {};
     }
--- a/src/app/shared/mock-data.service.ts
+++ b/src/app/shared/mock-data.service.ts
@@ -65,7 +65,7 @@ export class MockDataService {
 
   public createWorkItemOrEntity(extraPath: string, entity: any): any {
     if (extraPath && extraPath.indexOf('/') != -1) {
-      // request for subentities on a wi 
+      // request for subentities on a wi
       var parts = extraPath.split('/');
       var wiId = parts[0];
       var subselect = parts[1];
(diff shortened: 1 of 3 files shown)


src/app/shared/mock-data.service.ts
```diff
@@ -186,6 +192,15 @@ export class MockDataService {
     return this.makeCopy(this.workItemLinks);
   }
 
+  public getWorkItemLinksForWorkItem(workItemId: string): any {
+    var result: any = [];
+    for (var i = 0; i < this.workItemLinks.length; i++)
+      if (this.workItemLinks[i].relationships.source.data.id === workItemId) {
```
sanbornsen: Not necessarily current work item has to be in source, it can be in destination as well.

src/app/shared/mock-data.service.ts
```diff
@@ -186,6 +192,15 @@ export class MockDataService {
...
+        result.push(this.makeCopy(this.workItemLinks[i]));
```
sanbornsen: I saw how did you implement makeCopy, but that'll fail in some cases. I'd suggest to use cloneDeep from lodash.

src/app/shared/mock-data.service.ts
michaelkleinhenz: Mmmh..I discussed that with Naina. We came to the conclusion that only the source should be relevant. Are you sure? Because that would mean that when I am looking for the links for a work item, I would also get all links that are pointing "towards" that item, not only outgoing links..

src/app/shared/mock-data.service.ts
michaelkleinhenz: In which cases? I would suspect if it fails, then the data structure is weird anyway, because the serialization must be done at some point anyway (on the rest request for example). So it would be interesting to see in which cases it fails.

src/app/shared/mock-data.service.ts
sanbornsen: Yes, that is the case. Both forward and backward links are displayed right now.

michaelkleinhenz: Ok, will update that then.

sanbornsen:
Do not merge yet. Wait for this PR - https://github.com/michaelkleinhenz/fabric8-planner/pull/1 to get reviewed by @michaelkleinhenz  and merged.

michaelkleinhenz:
[test]

michaelkleinhenz:
[test]

michaelkleinhenz:
[test]

michaelkleinhenz:
[test]

PR #678, opened 2017-01-11
Remove unnecessary float property from work itme element

vikram-raj:
Fix https://github.com/fabric8io/fabric8-planner/issues/677

### DIFF
--- a/src/app/work-item/work-item-board/work-item-board.component.html
+++ b/src/app/work-item/work-item-board/work-item-board.component.html
@@ -1,20 +1,17 @@
 <div id="board_topWorkItems" class="container-fluid col-md-12 col-sm-12 col-xs-12 boardWrapper">
   <div class="board container-cards-pf">
-      <!-- board columns -->
-      <div class="board-lane" *ngFor='let lane of lanes'>
-
-        <section class="col-md-12 col-sm-12 col-xs-12 board-lane-column">
-          <div class="lane-header">
-            <span class="dib">{{lane.option}}</span>
-            <span class="issueCount dib" >{{workItems | almFilterBoardList:lane.option | count}}</span>
-          </div>
-          <div *ngFor='let item of workItems | almFilterBoardList:lane.option'>
+    <!-- board columns -->
+    <div class="board-lane" *ngFor='let lane of lanes'>
+      <section class="col-md-12 col-sm-12 col-xs-12 board-lane-column">
+        <div class="lane-header">
+          <span class="dib">{{lane.option}}</span>
+          <span class="issueCount dib" >{{workItems | almFilterBoardList:lane.option | count}}</span>
+        </div>
+        <div *ngFor='let item of workItems | almFilterBoardList:lane.option'>
           <div class="board-card" (click)="gotoDetail(item)">
             <aside>
-              <div class="pull-left">
-                <span class="wi-type-icon" almIcon [iconType]="item.relationships?.baseType?.data?.id"></span>
-              </div>
-              <span class="pull-left"> {{item.id}} </span>
+              <span class="wi-type-icon" almIcon [iconType]="item.relationships?.baseType?.data?.id"></span>
+              <span> {{item.id}} </span>
               <button class="pull-right btn btn-link dropdown-toggle color-grey" type="button">
                 <span class="fa fa-ellipsis-v"></span>
               </button>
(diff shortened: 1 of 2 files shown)


src/app/work-item/work-item-board/work-item-board.component.html
```diff
@@ -33,18 +32,18 @@
...
-                    src="{{item.relationalData.assignees[0]?.attributes.imageURL + '&s=20'}}"
-                    onError="this.src='https://avatars0.githubusercontent.com/u/563119?v=3&s=20'" />
+                     src="{{item.relationalData.assignees[0]?.attributes.imageURL + '&s=20'}}"
+                     onError="this.src='https://avatars0.githubusercontent.com/u/563119?v=3&s=20'" />
                 <span *ngIf="!item.relationalData?.assignees?.length" class="fa fa-user not-assigned-user-icon"></span>
               </div>
             </aside>
-
-           </div>
+            <div class="clear"></div>
```
sanbornsen: What is this for?

vikram-raj: It is for clearing the float So, that next element is not affected by previous element float property.

You can check all about float [here](https://css-tricks.com/all-about-floats/).

debloper:
- Not a big fan of clearfixes (not a fan of layouts needing to use clearfixes, that is). It's a hack that covers up (a particular kind of) layout issue, instead of resolving it. It made sense in old days when browser support for better layout mechanisms weren't as good.

- I know it took us a while to move away from table-based layouts to floating-grids. I'm hopeful it'll take us less time and resistance to move from floating-grids to flexboxes. Perhaps we should start preparing for that direction, as what we're building is for tomorrow, not for yesterday.

- However, **if you really need to use it here** (for the time being), reuse from existing declaration instead of creating another class. IIRC, bootstrap/pf and obsidian-generator both have **`clearfix`** mixins that you can readily use.

src/app/work-item/work-item-board/work-item-board.component.scss
```diff
@@ -151,11 +148,15 @@
   overflow-y: scroll;
 }
 
+.clear {
+  clear: both;
+}
+
```
debloper: No need to create another `.clear` class, please reuse the bootstrap CSS class for this.
It's defined in `node_modules/bootstrap/less/mixins/clearfix.less` and used in patternfly for your reference.

src/app/work-item/work-item-board/work-item-board.component.html
```diff
@@ -33,18 +30,18 @@
               </div>
               <div class="col-md-2 col-sm-2 col-xs-3 no-padding user-avatar">
                 <img *ngIf="item.relationalData?.assignees?.length"
-                    src="{{item.relationalData.assignees[0]?.attributes.imageURL + '&s=20'}}"
-                    onError="this.src='https://avatars0.githubusercontent.com/u/563119?v=3&s=20'" />
+                     src="{{item.relationalData.assignees[0]?.attributes.imageURL + '&s=20'}}"
+                     onError="this.src='https://avatars0.githubusercontent.com/u/563119?v=3&s=20'" />
                 <span *ngIf="!item.relationalData?.assignees?.length" class="fa fa-user not-assigned-user-icon"></span>
               </div>
+              <div class="clear"></div>
```
debloper: Using bootstrap's clearfix, you maybe able to avoid using extra div.

src/app/work-item/work-item-board/work-item-board.component.html
```diff
@@ -1,20 +1,17 @@
...
-          <div class="lane-header">
-            <span class="dib">{{lane.option}}</span>
-            <span class="issueCount dib" >{{workItems | almFilterBoardList:lane.option | count}}</span>
-          </div>
-          <div *ngFor='let item of workItems | almFilterBoardList:lane.option'>
+    <!-- board columns -->
+    <div class="board-lane" *ngFor='let lane of lanes'>
+      <section class="col-md-12 col-sm-12 col-xs-12 board-lane-column">
+        <div class="lane-header">
+          <span class="dib">{{lane.option}}</span>
```
nimishamukherjee: Is the dib class needed? I removed it and the UI looked fine.

vikram-raj: Yes, it's not needed. Because it is a span element and no need to make it inline-block until we need some top or bottom padding or margin.

SMahil: lane.option needs to aligned center. For that reason dib is given

Issue #643, opened 2017-01-06
In the detailed Workitem creation dialog, a status of "Created" is always displayed - even if the workitem is not yet created


ldimaggi:
See attached screenshot:

![screenshot from 2017-01-06 12-48-08](https://cloud.githubusercontent.com/assets/642621/21727648/5064e340-d410-11e6-8030-47655a9dce22.png)

nimishamukherjee:
@fabric8io/uxd confirming the Created behaviour - we will not display the "Created" text during work item creation via the dialogue box.

nimishamukherjee:
@ldimaggi / @naina-verma please verify the fix on demo.almighty.

naina-verma:
@nimishamukherjee , Fix works fine !

PR #681, opened 2017-01-11
Remove Created field when adding a new work item using the dialogue box (#643)

nimishamukherjee:
Remove Created field when adding a new work item using the dialogue box #643


nimishamukherjee:
[test]

Issue #736, opened 2017-01-13
Update UI of Dialog base creation of WI 


SMahil:
Update UI of Dialog base creation of WI  as per the Visuals

Visuals:  https://redhat.invisionapp.com/share/HJ9YNYX2T#/screens

PR #751, opened 2017-01-16
match UI for WI type modal

SMahil:
#736 

- Match UI for WI type modal
- Match classes for WI action bar as per Patternfly

### DIFF
--- a/src/app/work-item/work-item.component.html
+++ b/src/app/work-item/work-item.component.html
@@ -1,83 +1,100 @@
 <div class="work-item-page">
-
+  <!-- select WI type page -->
   <div *ngIf="loggedIn && showTypesOptions" class="types-modal">
-    <h5 class="col-md-12 col-sm-12 col-xs-12">
-      <span class="pull-left">Click to select a type </span>
-      <span class="pficon-close detail-close pull-right" (click)="closePanel()"></span>
-    </h5>
-    <ul class="wiTypeList col-md-12 col-sm-12 col-xs-12 grid">
-      <li *ngFor="let type of workItemTypes" (click)="onChangeType(type.name)">
-        <div class="wiTypeListBlk">
-          <i almIcon [iconType]="type.name" class="xl-font"></i>
-          <p>{{type.name}}</p>
+    <div class="container-fluid container-cards-pf">
+      <div class="row row-cards-pf">
+        <div class="col-md-12 col-sm-12 col-xs-12 no-padding">
+          <h2 class="pull-left">Choose a new work item type </h2>
+          <i class="pficon-close detail-close pull-right" (click)="closePanel()"></i>
         </div>
-      </li>
-    </ul>
-  </div>
-
-  <div class="work-item-action-bar">
-    <div class="list-action">
-
-      <!--Move a work item up / down -->
-      <div class="dropdown move-dropdown" dropdown>
-        <button class="btn btn-default dropdown-toggle" type="button" id="wi_filter_dropdown" data-toggle="dropdown" dropdown-open>
-          Move
-          <span class="caret"></span>
-        </button>
-        <ul class="dropdown-menu dropdown-menu-left" role="menu" aria-labelledby="wi_filter_dropdown">
-          <li>
-            <a role="menuitem" tabindex="-1" (click)="moveItem('top')">Move to Top</a>
-          </li>
-          <li>
-            <a role="menuitem" tabindex="-1" (click)="moveItem('bottom')">Move to Bottom</a>
-          </li>
-          <li role="presentation" class="divider"></li>
-          <li>
-            <a role="menuitem" tabindex="-1" (click)="moveItem('up')">Move Up</a>
-          </li>
-          <li>
-            <a role="menuitem" tabindex="-1" (click)="moveItem('down')">Move Down</a>
-          </li>
-        </ul>
+        <div *ngFor="let type of workItemTypes" class="col-xs-12 col-sm-6 col-md-4 col-lg-3" (click)="onChangeType(type.name)">
+          <div class="card-pf card-pf-view card-pf-view-select">
+          <div class="card-pf-body">
+            <div class="card-pf-top-element">
+              <i almIcon [iconType]="type.name" class="card-pf-icon-circle"></i>
+            </div>
+            <h2 class="card-pf-title text-center">
+              {{type.name}}
+            </h2>
+            <p class="card-pf-info">Definition goes here. Lorem Ipsum dolor sit amet, consectetur adpisicing</p>
+          </div>
+          </div>
+        </div>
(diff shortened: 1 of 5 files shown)


sanbornsen:
This PR has changes other than just WI type modal. Please rename the PR, commit and list down all the changes in commit message and PR description as well.

src/app/work-item/work-item.component.html
```diff
@@ -1,83 +1,99 @@
...
-            <span>No filters available</span>
-          </li>
-          <li *ngFor="let filter of filters;" role="presentation" (click)="activeFilter(filter.id)">
-            <a role="menuitem" tabindex="-1">{{filter.name}}</a>
-          </li>
-        </ul>
+    <div class="container-fluid container-cards-pf">
+      <div class="row row-cards-pf">
+      <div class="col-md-12 col-sm-12 col-xs-12 no-padding">
+        <h2 class="pull-left">Click to select a type </h2>
```
nimishamukherjee: As per the wireframe the title needs to be 'Choose a new work item type'

src/app/work-item/work-item.component.html
```diff
@@ -1,83 +1,99 @@
...
-          (click)="showTypes()">
-        <span class="fa fa-plus icon"></span>
+        <div *ngFor="let type of workItemTypes" class="col-xs-12 col-sm-6 col-md-4 col-lg-3" (click)="onChangeType(type.name)">
+          <div class="card-pf card-pf-view card-pf-view-select">
+          <div class="card-pf-body">
+            <div class="card-pf-top-element">
+              <i almIcon [iconType]="type.name" class="card-pf-icon-circle"></i>
+            </div>
+            <h2 class="card-pf-title text-center">
+              {{type.name.substring(7, type.length)}}
```
debloper: Use `type.name.split(".")[1]` instead for better performance and abstraction.

**Performance:** It's better to delimit and cut the string once than doing 2 looks-ups (`substring` args). JS is not good at string processing. Leaner the better.

**Abstraction:** If the `system` changes to `systems` or `types` or something minute, the code will need no change, as long as the delimiter is same. It just wants the right hand side after the dot.

src/app/work-item/work-item.component.scss
```diff
@@ -8,93 +8,65 @@
...
       }
 
       .add-detailed-wi {
-        font-size: em(20);
-        width: em(26);
-        height: em(26);
-        background: $color-pf-blue-400;
-        color: $color-pf-white;
+        width: em(17);
+        height: em(17);
```
debloper: Why not `em(14)` as per UXD? Any particular reason for 17px?
BTW, as per UX guideline, it has to be a 40x40 active element.

src/assets/stylesheets/custom.scss
```diff
@@ -138,8 +136,39 @@ $textCenter: center;
...
 
-//user -gravatar
+.card-pf-view {
+  .card-pf-top-element {
+    .card-pf-icon-circle {
+      @include Border(em(0.8), solid, $color-pf-blue-300); //#39a5dc;
+      @include borderRadius(50%);
+      @extend .db;
+      font-size: em(36);
+      height: em(34); //2.84em
```
debloper: Marginally misses the UXD spec. it's 100x100px including border.
UXD specs it to be 102x102px (unsure with or without border).

I suggest apply em(36) to all, that'll *likely* match the spec.

src/app/work-item/work-item.component.scss
SMahil: reason behind using em(17) is whole action bar has different styling and it limits the exact css styles given for icon.

src/assets/stylesheets/custom.scss
SMahil: em(36) doesnot give 102px here as the font-size acts as base font for the div

src/app/work-item/work-item.component.html
SMahil: @vikram-raj @sanbornsen @nimishamukherjee can anyone confirm this change as it was already there

nimishamukherjee: @SMahil please go ahead and change it as per @debloper's comment.

src/app/work-item/work-item.component.scss
debloper:
```css
.add-detailed-wi {
  width: em(14);
  height: em(14);
  ...
}
```
This at least matches the icon size without any side effect.

I understand the pf-toolbar is a mess, beyond your current scope (quite some custom style override has to be added to fix it) but it isn't something we can skip for long (even if we do in this patch). It eventually needs fixing.

src/assets/stylesheets/custom.scss
debloper: We should bounce it back against @fabric8io/uxd team if they'd prefer a static `px` value instead of `em` here. From what I understand, serving it with different form factor isn't a choice here. Do we need to force `em`? Although, same can be asked about why 102px has been prescribed to be precise, but well...

(To clarify, I'm grossly fine with the current implementation, it's not hindering usability IMO in any way, but I don't like the inconsistency with the UXD spec vs implementation. Either we need to fix/justify the spec, or follow it properly.)

src/assets/stylesheets/custom.scss
```diff
@@ -138,8 +136,39 @@ $textCenter: center;
...
+      width: em(34);
+      line-height: em(34);
+      margin:0 auto;
```
debloper: nitpick: please add a space before the `0`

src/app/work-item/work-item.component.html
```diff
@@ -1,83 +1,99 @@
...
-          (click)="showTypes()">
-        <span class="fa fa-plus icon"></span>
+        <div *ngFor="let type of workItemTypes" class="col-xs-12 col-sm-6 col-md-4 col-lg-3" (click)="onChangeType(type.name)">
+          <div class="card-pf card-pf-view card-pf-view-select">
+          <div class="card-pf-body">
+            <div class="card-pf-top-element">
+              <i almIcon [iconType]="type.name" class="card-pf-icon-circle"></i>
+            </div>
+            <h2 class="card-pf-title text-center">
+              {{type.name.split(".")[1]}}
```
vikram-raj: Need to change this to {{ type.name }}.  Now you can get WI type directly by type.name.

https://github.com/fabric8io/fabric8-planner/pull/782

src/assets/stylesheets/custom.scss
SMahil: This become a separate Issue. Will fix it in new PR

debloper: Cool then, good for now.

SMahil:
[test]

SMahil:
[test]

Issue #758, opened 2017-01-16
Invalid configuration of multiple routes


ppitonak:
Some routes contain configuration like this:

```
const routes: Routes = [ 
  {
    path: 'chat',
    component: ChatComponent,
    children: [
      { path: '' }
    ]
  },
];
```

Empty path is not a valid configuration. Although it works fine with Angular 2.1.1/Angular Router 3.1.1, it fails with Angular 2.3.0 with following errors:

```
Invalid configuration of route 'chat/'. One of the following must be provided: component, redirectTo, children or loadChildren
```

More info in https://github.com/angular/angular/issues/12827 and https://github.com/angular/angular/pull/13224

We need to fix it in order to upgrade to Angular >= 2.3.0

PR #759, opened 2017-01-16
empty paths in router configuration fixed (#758)

ppitonak:
This is a pre-requisite for upgrading to newer versions of Angular

### DIFF
--- a/src/app/chat/chat-routing.module.ts
+++ b/src/app/chat/chat-routing.module.ts
@@ -6,12 +6,7 @@ import { ChatComponent } from './chat.component';
 const routes: Routes = [
   {
     path: 'chat',
-    component: ChatComponent,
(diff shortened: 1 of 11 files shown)


sanbornsen:
@ppitonak These are for references. In case you want to add children and you don't define an empty path, that URL will not work without any child URL. For example, if you have two children of **/parent** as **/parent/child1** and **parent/child2**, and if you don't mention an empty path in parents' children section then, **/parent** will not work but **/parent/child1** and **parent/child2** will. However, we are going to remove all those modules from planner and they are eventually gonna go into platform.

ppitonak:
@sanbornsen you are right, my impl doesn't work correctly, thanks for explaining. I'll close this PR and leave issue open.

PR #764, opened 2017-01-17
Added test for UI operation to change order of execution - test is mo…

ldimaggi:
…stly commented out at present due to bug in PhantomJS - but it runs cleanly with Firefox and Chrome. We can uncomment the test when we move the UI tests to runo nFirefox and Chrome in Centos CI.


naina-verma:
@ldimaggi ,LGTM

nimishamukherjee:
[test]

PR #768, opened 2017-01-17
Few negative cases covered for previous features

naina-verma:
Test covered 
- Verify that workitems cannot be assigned to non-existent users
- Verify that the title field must not be unique
- Verify that the description field can have its contents deleted
- Check before while adding description the field should not display “undefine/null’
- Verify the HTML tags should be treated as text while writing description/title.
- Test description/title can be in other languages also acceptable
- Verify that all fields are read-only for these users 
- Verify that data is persisted and is not truncated if text fields receive data with a length greater that 255 characters.
- Verify visual display of data with a length greater that 255 characters.
- Verify if the date/time is in future it should say “in X time “

### DIFF
--- a/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
+++ b/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
@@ -276,6 +276,9 @@ class WorkItemListPage {
   get filterDropdown () {
     return  element(by.id("wi_filter_dropdown"));
   }
+  filterDropdownId () {
+    return  element(by.id("wi_filter_dropdown"));
+  }
 
   clickFilterDropdown () {
     return this.filterDropdown.click();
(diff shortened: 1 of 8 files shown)


src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
```diff
@@ -272,6 +272,9 @@ class WorkItemListPage {
   get filterDropdown () {
     return  element(by.id("wi_filter_dropdown"));
   }
+  filterDropdownId () {
```
ldimaggi: This code is correct - but - why do we need 2 functions that access the same UI element?

naina-verma: This for just another way to test by different approach , Well when you manually go to Link item everything is clickable , So i have added multiple methods to verify that scenario!

ldimaggi: OK - sounds good - approved!

nimishamukherjee:
[test]

PR #776, opened 2017-01-18
Added new test for changing order of execution of workitems through m…

ldimaggi:
…ove options provided in a workitem's "kebab" menu, added blocked (and commneted out) drag and drop test (test is blocked by bug in protractor and its interaction with browsers and angular2).


nimishamukherjee:
[test]

PR #782, opened 2017-01-19
Fix display of names in work item create dialog

kwk:
As of https://github.com/almighty/almighty-core/commit/7a5b4fd32588eb7304bea7a72746f0894bae1cf5 we no longer store WIT names with the `"system."` prefix. Hence, we must not strip-off 7 characters in  front of WIT names when displaying.

When you click this button in the top right corner 
bildschirmfoto von 2017-01-19 23-35-53

you get this without this change:
bildschirmfoto von 2017-01-19 23-35-37

Quite obviously somebody removed the first 7 characters and not just left trims any occurence of "system.".

### DIFF
--- a/src/app/work-item/work-item.component.html
+++ b/src/app/work-item/work-item.component.html
@@ -9,7 +9,7 @@
       <li *ngFor="let type of workItemTypes" (click)="onChangeType(type.name)">
         <div class="wiTypeListBlk">
           <i almIcon [iconType]="type.name" class="xl-font"></i>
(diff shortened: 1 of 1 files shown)


nimishamukherjee:
[test]

nimishamukherjee:
Created #783. Quick add fails as it submits system.workitemtype. Remove 'system.' prefix from the quick add work item form.

kwk:
Checking that fix is correctly deployed:

bildschirmfoto von 2017-01-22 05-17-22

PR #786, opened 2017-01-20
Update iteration added with unit test (fixes #710)

### LINKED ISSUE #710: Write service to PATCH an iteration content
This task is part of #647

sanbornsen:


src/app/iteration/iteration.service.ts
```diff
@@ -80,10 +81,35 @@ export class IterationService {
...
    */
-  getIterationList() {
-    return this.iterations;
+  updateIteration(iteration: IterationModel): Promise<IterationModel> {
+    return this.http
+      .patch(iteration.links.self, { headers: this.headers })
+      .toPromise()
+      .then (response => {
+        if (response.status != 200) {
+          throw new Error('API error occured');
```
bartoszmajsak: I'm not sure how backend looks like - but generally speaking shouldn't we only treat 4xx and 5xx as errors? I'm also not sure if 200 is a correct return code for PATCH - shouldn't it be 204 for example?

sanbornsen: @bartoszmajsak Angular's http handles that smartly. I believe, error from live API goes to catch part by default. This block is only for unit test purpose where we are mocking the response. I see in mock response even if the status is 4xx or 5xx it doesn't go to catch section. 
Though, your point is totally valid. I should just check for error codes instead of != 200. I'll update that.

src/app/iteration/iteration.service.ts
```diff
@@ -1,11 +1,12 @@
+import { cloneDeep } from 'lodash';
 import { Http, Headers } from '@angular/http';
 import { AuthenticationService } from './../auth/authentication.service';
 import { IterationModel } from './../models/iteration.model';
 import { Injectable } from '@angular/core';
 
 @Injectable()
 export class IterationService {
-  private iterations: IterationModel[] = [];
+  public iterations: IterationModel[] = [];
```
nimishamukherjee: Why public?

sanbornsen: Because we'll need the reference all over the component.

src/app/iteration/iteration.service.ts
bartoszmajsak: I think exposing something for unit test purposes, or adjusting the production code for that is a smell. There should be a very strong reason to really go this way.

src/app/iteration/iteration.service.ts
```diff
@@ -80,10 +81,35 @@ export class IterationService {
...
+   * @return updated iteration's reference from the list
    */
-  getIterationList() {
-    return this.iterations;
+  updateIteration(iteration: IterationModel): Promise<IterationModel> {
+    return this.http
+      .patch(iteration.links.self, { headers: this.headers })
+      .toPromise()
+      .then (response => {
+        if (/^[5, 4][0-9]/.test(response.status.toString())) {
```
bartoszmajsak: Wouldn't it be just better to check if `response.status >= 400 && response.status <= 599`? regex is not for free I guess. On top of that this code is duplicated above - maybe small extract?

Don't we have some meaningful explanation of what was the error in the `response`? If so maybe it's good to populate it as a part of `new Error(....)`?

sanbornsen: As I explained earlier, this check and error is only for testing. The mock http error response is not leading the promise to catch section. I can't figure out why it's not happening. So, if it's a live error from API and it has a meaningful message it goes to catch section by default and it'll never come to this check.
What do you mean by "regex is not for free" ?

bartoszmajsak: I meant it does not make sense here to involve regex which is more expensive operation that just simple number comparison. Also the condition is more readable from my point of view - you immediately see what is the logic about.

sanbornsen: @bartoszmajsak I see what you mean. We'll collaborate with backend and get exact error code and update this module.

PR #793, opened 2017-01-21
Added new test for filtering workitems - and supporting change to moc…

ldimaggi:
…k framework.

### DIFF
--- a/src/app/shared/mock-data.service.ts
+++ b/src/app/shared/mock-data.service.ts
@@ -59,6 +59,25 @@ export class MockDataService {
 
   // work items and dependend entities (comments, ..)
 
+  /* Filter the workitems based on assignee */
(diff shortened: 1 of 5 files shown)


protractor.config.js
```diff
@@ -4,7 +4,7 @@ exports.config = {
     seleniumAddress: 'http://localhost:4444/wd/hub',
     specs: ['src/tests/**/*.spec.js'],
     jasmineNodeOpts: {
-        defaultTimeoutInterval: 60000
+        defaultTimeoutInterval: 30000
```
sanbornsen: We purposely increased the timeout to support long set of test. We were having random test failure otherwise.
@naina-verma please comment.

ldimaggi: Changed the timeout to 60 sec - is this what you wanted?

PR #795, opened 2017-01-23
Move iteration HTML to iteration module.

vikram-raj:
fix: https://github.com/fabric8io/fabric8-planner/issues/794


src/app/work-item/work-item.component.html
```diff
@@ -80,197 +80,8 @@ <h5 class="col-md-12 col-sm-12 col-xs-12">
...
-                </div>
-              </li>
-            </ul>
-          </div>
-        </div>
-        <div *ngIf="loggedIn" class="create-iteration-bar pointer">
-          <span><i class="pficon-add-circle-o fa-icon"></i>
-          <span class="create-iteration-label">Create Iteration</span></span>
-        </div>
+      <iteration></iteration>
```
sanbornsen: Component should be named with a prefix. So far we have been using **alm** but since it's not **alm** anymore, let's use **fabric8-planner**. The name should be **fabric8-planner-iteration**.
I am not sure if it supports number or not, please try and check.

src/app/work-item/work-item.component.scss
```diff
@@ -128,87 +128,14 @@
...
-        flex: auto;
-        overflow: auto;
-      }
-      .badge-wrapper {
-        display: flex;
-        flex-direction: column;
-        justify-content: center;
-      }
-      .iteration-panel-header {
+      iteration {
```
sanbornsen: Use a class name instead. See if that work.

src/app/work-item/work-item.component.html
vikram-raj: @sanbornsen component name not support number.

sanbornsen: @vikram-raj then do it, **fab-planner-iteration**

nimishamukherjee:
[test]

Issue #810, opened 2017-01-24
Iteration and workItem detail overlapped ,Distorted display 


naina-verma:
Iteration and workItem detail overlapped ,Distorted display 
![screenshot from 2017-01-24 11-33-37](https://cloud.githubusercontent.com/assets/19172870/22236042/197d6934-e229-11e6-936a-91b5a0cfc8d8.png)

mindreeper2420:
@vikram-raj @naina-verma why is UX needed for this? This is a z-index bug. Looking at the code, the panel is set to a crazy high 9999.

PR #812, opened 2017-01-24
fix overlapped issue of create Iteration modal and workItem detail.

vikram-raj:
- Reduce the z-index of detail panel.
Fix: #810

### DIFF
--- a/src/app/work-item/work-item-detail/work-item-detail.component.scss
+++ b/src/app/work-item/work-item-detail/work-item-detail.component.scss
@@ -253,7 +253,7 @@
   bottom: 0;
   width: 50%;
   min-width: em(250);
-  z-index: 9999;
+  z-index: 100;
   padding: em(20) 0;
   border-left: 1px solid darken($color-pf-white, 20%);
   overflow-y: scroll;


debloper:
Seems like a fix for now, but we'll hit a point where unknown order/precedence of z-index of multiple elements will cause regressions.

I suggest having a z-index order/precedence range guide of our UI, and get it reviewed by @fabric8io/uxd. Knowing that a modal can have 1000-1200 z-index, a floating button or something can have 500-600 z-index and a sliding sidebar can have 100-200 z-index etc (all these are for e.g. only, please don't nitpick on values) will make a huge difference on how many complicated regressions we will avoid in later stages of the app.

vikram-raj:
@debloper yes, I agree with you. Need a feedback from @fabric8io/uxd.

nimishamukherjee:
@vikram-raj please create an issue on https://github.com/fabric8io/fabric8-ux for this (which element would sit on top of which)?

mindreeper2420:
Questions:
- why does the panel have a z-index of 100? Why so high? Also, I was unable to find a z-index for any other element - modal, navigation, or anything else.
- Do other elements already have z-indexes?
- Is Bootstrap / PatternFly being used to control the layering, as it should? Those frameworks cover layering for modals, slideouts, and other aspects. Custom code should not be there for that. @michaelkleinhenz, @michaelvirgil - how much of Bootstrap/PatternFly is being used at this point?

We will most likely not be able to give you a z-index for every single element - that is what the embedded frameworks are for. The CSS for each framework will properly handle the button layering, modal layering, etc.

vikram-raj:
- z-index property only affects elements that have position value other than static. And modal have z-index 1050. which is set by bootstrap and to make sure it will be always on top.

- Not all elements have z-index property. Z-index property set to only those elements which have position fixed or absolute property.

- Bootstrap / Patternfly control the layering by giving z-index property.

@mindreeper2420  Here we just wanted to know the order of layering of elements. (modal, detail page, kabal menu.)

mindreeper2420:
@vikram-raj - As you stated, Bootstrap / PatternFly control the layering by giving the z-index property. For anything not covered by either of those libraries, please use what makes logical sense - i.e. the kabob should not appear above the detail panel.

vikram-raj:
Ok @mindreeper2420. Thanks.

nimishamukherjee:
[test]

PR #813, opened 2017-01-24
Iteration page object model and few test

naina-verma:
Iteration page object model and few test
- Verification of elements are in place 
- Verification set the title and description
- Verify  help box / dialog title is visible

### DIFF
--- a/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
+++ b/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
@@ -433,12 +433,6 @@ class WorkItemListPage {
   clickWorkItemMovePulldownDown (parentElement) {
     return this.workItemMovePulldownDown(parentElement).click();
   }
-
-  /* Workitem filter pulldown */
-  workItemFilterPulldown () {
-    return workItemElement.element(by.css(".dropdown.filter-dropdown"));
-  }
-
   /* Access the Kebab 'move to top' element relative to its parent workitem */
   workItemKebabMoveToTopButton (parentElement) {
     browser.wait(until.presenceOf(parentElement.element(by.css('.workItemList_MoveTop'))), constants.WAIT, 'Failed to find clickWorkItemKebabButton');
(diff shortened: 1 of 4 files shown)


naina-verma:
[test]

src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
```diff
@@ -276,13 +276,37 @@ class WorkItemListPage {
   get filterDropdown () {
     return  element(by.id("wi_filter_dropdown"));
   }
-  filterDropdownId () {
```
ldimaggi: So - this is not needed? (Was it ever used?) Thx!

src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
```diff
@@ -432,8 +450,42 @@ class WorkItemListPage {
...
+    return this.iterationTitle.sendKeys(newTitleString);
+  }
+  get iterationDescription  (){
+    return element(by.id("iteration-description")).click();
+  }
+  setIterationDescription  (newString,append){
+     if (!append) { this.iterationDescription.clear(newString) };
+    return this.iterationDescription.sendKeys(newString);
+  }
+  clickCreateIteration  (){
```
ldimaggi: We should have separate functions to define the create and cancel buttons, and to click the buttons. Having the 2 buttons will keep the page object model elements consistent. Thx!

src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
naina-verma: I discovered this duplicated function during this PR , it looks like never used so removed it :)

src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
naina-verma: I  did not make separate functions to define button and directly used them, bcuz I  couldn't think of any other use case rather then clicking on it for now :). And yes we can use the general method for checking the presence of the elements. 

I will add general definition in different methods.

naina-verma:
[test]

naina-verma:
[test]

naina-verma:
[test]

naina-verma:
[test]

PR #816, opened 2017-01-24
Added test for troublesome strings (unicode, multi-byte, etc.) - refe…

ldimaggi:
…rence strings defined here: https://github.com/minimaxir/big-list-of-naughty-strings - Note that the tests run error-free on PhantomJS, but fail with a Script Injection error on Chrome and Firefox. This failure is tracked here:  https://github.com/fabric8io/fabric8-planner/issues/807


sanbornsen:
[test]

PR #817, opened 2017-01-25
Fetching and listing iteration (fixes #708)

### LINKED ISSUE #708: Display list of iterations
This task is part of #647

sanbornsen:
For now it fetches the list of spaces and work with the first space in the list. Once we implement the shared service or anything similar to fetch the default space this temporary code for fetching spaces will be removed.

Space and iteration need to be created from the backend or using the API to test this feature.

### DIFF
--- a/src/app/iteration/iteration.component.html
+++ b/src/app/iteration/iteration.component.html
@@ -21,10 +21,10 @@
       </div>
     </div>
     <ul class="iteration-expanded-div" [collapse]="isCollapsedCurrentIteration">
-      <li class="active">
+      <li class="active" *ngFor="let iteration of currentIterations">
         <div class="iteration-subheader">
           <div>
-            <strong>Iteration 1</strong>
+            <strong>{{iteration.attributes.name}}</strong>
           </div>
           <div class="dropdown dropdown-kebab-pf" dropdown>
             <button class="btn btn-link dropdown-toggle"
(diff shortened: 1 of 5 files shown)


src/app/iteration/iteration.component.html
```diff
@@ -68,7 +68,7 @@
                tooltip="{{(100 - barchatValue) + '%'}}">
           </div>
         </div>
-        Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+        <!-- Lorem Ipsum is simply dummy text of the printing and typesetting industry. -->
```
vikram-raj: Don't we have description text now?

src/app/iteration/iteration.module.ts
```diff
@@ -1,3 +1,4 @@
+import { IterationService } from './iteration.service';
```
vikram-raj: This import should be after angular imports.

src/app/iteration/iteration.component.html
sanbornsen: Not right now

src/app/iteration/iteration.module.ts
sanbornsen: Done

src/app/iteration/iteration.service.ts
```diff
@@ -130,4 +130,21 @@ export class IterationService {
...
+
+  // Temporary space fetch
+  getSpaces() {
+    return this.http
+    .get(process.env.API_URL + 'spaces', { headers: this.headers })
+    .toPromise()
+    .then((response) => {
+      let spaces = response.json().data;
+      if (spaces.length) {
+        return spaces[0];
```
debloper: I understand this is temporary fetch, but why are we only extracting/using the 1st space only?

PR #866, opened 2017-02-01
POC for Parallel execution of functional tests

naina-verma:
This POC for Parallel execution of functional tests
Features : 
- This PR has capability to run functional tests in approximately 3 mins (max) 
- It is for  POC for parallel execution 
How it works 
- This PR does not trigger ./run_fuctinaltest.sh you need to run ./run_funct*.sh file manually, (testing Purpose, so that it will not affect master )
- Moved  those pages which are not implemented say "setting.spec.js , dashboard , pipleline " etc.. to TBD folder .
Will move them into other Suite (future scope)..

### DIFF
--- a/protractor.config.js
+++ b/protractor.config.js
@@ -3,6 +3,7 @@ exports.config = {
     getPageTimeout: 30000,
     seleniumAddress: 'http://localhost:4444/wd/hub',
     specs: ['src/tests/**/*.spec.js'],
+    exclude: ['src/tests/**/TBD/*.spec.js'],
     jasmineNodeOpts: {
         defaultTimeoutInterval: 60000
     },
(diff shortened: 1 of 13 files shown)


naina-verma:
Do not merge this PR , this is for testing purpose , i need to update it !

--Thanks

michaelkleinhenz:
Ok, so the parallel run is triggered by the config change, right? If that poses problems with the "shared test files", we could also run the tests in parallel using the build system. We need to switch to gulp anyway pretty soon (when we only do the library planner version). Gulp has the capability of running tasks in parallel with a sync point at the end. We could just run the tests using that as well.

naina-verma:
@michaelkleinhenz , Yes we can run tests in parallel for now using this file . Am waiting for 
tests to be fixed on Docker image first then i can run and update this PR accordingly .

naina-verma:
Hi @ldimaggi ,

I have made changes into the POC waiting for :https://github.com/fabric8io/fabric8-planner/pull/912 to be merged and test it back to back on CI/CD . Please review it.

PR #870, opened 2017-02-02
Add a pipe for multiline truncate. (fixes #868)

### LINKED ISSUE #868: Distorted Layout  "Iteration" Description
Distorted Layout  "Iteration" Description, displayed in Current Iterations
Expected : There should be content wrap for showing minimal display 

screenshot from 2017-02-02 09-19-51

vikram-raj:
fix: #868

### DIFF
--- a/src/app/pipes/alm-multiline-truncate.pipe.ts
+++ b/src/app/pipes/alm-multiline-truncate.pipe.ts
@@ -1,0 +1,35 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+/*
+ * Usage:
+ *   {{ aParagraph | almMultilineTruncate:wordCount }}
+ *
+ * Example:
+ *  paragraph = "Lorem Ipsum is simply dummy text of the printing and
+ *               typesetting industry. Lorem Ipsum has been the industry's
+ *               standard dummy text ever since the 1500s, when an unknown
+ *               printer took a galley of type and scrambled it to make a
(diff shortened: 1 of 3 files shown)


SMahil:
We already have multiline truncation in css. Do we need pipe?

src/app/pipes/alm-multiline-truncate.pipe.ts
```diff
@@ -0,0 +1,35 @@
...
+ *  industry. Lorem Ipsum has been the industry's standard dummy..."
+*/
+
+@Pipe({ name: 'almMultilineTruncate' })
+export class AlmMultilineTruncate implements PipeTransform {
+  transform(val: string, wordCount: number): any {
+    if (!val) {
+      return '';
+    }
+    let words = val.split(' ');
```
sanbornsen: Sometimes words are bigger, and 20 words might not fit in. Why don't we do it with width instead of word count with CSS?

vikram-raj:
@SMahil But that not work as expected.

SMahil:
@vikram-raj Can we check that tomorrow?

vikram-raj:
@SMahil Yes sure.

debloper:
1. Can we not use CSS for this?
2. If we really need the pipe, can you please change the prefix to `fab` instead of `alm`?

vikram-raj:
This issue is resolved in PR #925 using CSS.

PR #904, opened 2017-02-06
Silently ignore unknown creator/assignees

aslakknutsen:
NOTE: This should technically not happen, but it does currently due to a mix of old data and import of remote work items not doing 100% what they should. At least this allow the UI to display them.


nimishamukherjee:
Closing this PR as the production DB was cleared.

joshuawilson:
@aslakknutsen I saw you mention this PR. It looks like it should have been merged, yes?

@nimishamukherjee why didn't it get merged? It looks like a reasonable safety check.

aslakknutsen:
@joshuawilson The original reason for this has been fixed by a clean db, but it's still currently required if we use remote work item imports as UI currently doesn't separate between Users and Identity. e.g. a User is a known user via our KC, an Identity is 'anything' from any remote system which could be linked to a User but might not be. So Creator/ Assignee could either be a User or an Identity. But generally yes, it's not a bad safety. 

Generally, defensive programming is a good thing.

PR #912, opened 2017-02-07
Added space API to mock http API list

sanbornsen:
We have introduced a temporary space service, but we didn't have any mock service for that. Inside the docker container, being unable to reach the space endpoint, pages were failed to be loaded thus the tests were failing. 

I just have added the space endpoint to the mock service and used the mock service in iteration service (this is something that I missed while creating the service). Now the functional tests inside docker run fine.

### DIFF
--- a/src/app/shared/mock-http.ts
+++ b/src/app/shared/mock-http.ts
@@ -16,15 +16,15 @@ import { MockDataService } from './mock-data.service';
  * the stock http service when the mock is used. Currently, the service is
  * replaced 'manually'. When the Angular 2 API on the injectors are more stable,
  * this class should be submitted to the Angular injector service to replace
- * the http service on injection. Currently, this API is unstable and changes 
- * frequently, so we did not use it. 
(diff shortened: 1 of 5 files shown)


src/app/shared/mock-http.ts
```diff
@@ -182,6 +182,8 @@ export class MockHttp extends Http {
           return this.createResponse(url.toString(), 200, 'ok', this.mockDataService.getWorkItemLinkTypes() );
         case '/workitemlinkcategories':
           return this.createResponse(url.toString(), 500, 'not supported yet.', { } );
+        case '/spaces':
```
sanbornsen: Ignore all the other changes in this file but this.

naina-verma:
@sanbornsen , I think you need remove comment   from 
`# docker exec almighty-ui-builder ./run_functional_tests.sh` from cico_run_test.sh

sanbornsen:
[test]

PR #919, opened 2017-02-08
Implements iteration association from list view through modal dialog, fixes #714, fixes #716

### LINKED ISSUE #714: Create the modal dialog for iteration association of work items
- Populates interaction modal as per UXD

Wireframes: https://redhat.invisionapp.com/share/KA9CAYL7M#/screens/207562836
Visual design: https://redhat.invisionapp.com/share/FR9R8G6B9#/screens/211978573

Part of userstory: #648

### LINKED ISSUE #716: Service integration to update a work item’s iteration (read, update, remove iteration)
- Allow populating and editing WI's iterations from list view kebab menu
- Allow populating and editing WI's iterations from detail view dropdown

Part of userstory: #648

debloper:
Thanks @vikram-raj for modularizing the modal component.


src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  onChangeIteration(iteration: IterationModel): void {
+    this.selectedIteration = iteration;
+    this.dropdownButton.nativeElement.innerHTML = this.selectedIteration.attributes.name + ' <span class="caret"></span>'
+  }
+
+  assignIteration(event: MouseEvent): void {
+    this.workItem.relationships.iteration = {
+      data: {
+        id: this.selectedIteration.id,
```
vikram-raj: If user not select any iteration from the dropdown and click the associate button then it will throw error id of property undefined.
You need to handle this case.

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    this.iterations = this.iterationService.iterations;
+  }
+
+  ngOnChanges() {
+
+  }
+
+  onChangeIteration(iteration: IterationModel): void {
+    this.selectedIteration = iteration;
+    this.dropdownButton.nativeElement.innerHTML = this.selectedIteration.attributes.name + ' <span class="caret"></span>'
```
vikram-raj: semicolon missing.

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
debloper: Good catch

debloper: Nice catch, fixed in later commit.

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  onChangeIteration(iteration: IterationModel): void {
+    // Cache changed iteration
+    this.selectedIteration = iteration;
+
+    // Update dropdown button content
+    this.dropdownButton.nativeElement.innerHTML = this.selectedIteration.attributes.name + ' <span class="caret"></span>';
+
+    // Enable/disable confirmation button if iteration was changed
+    let commit = document.querySelector("#associate-iteration-button");
```
sanbornsen: This is not Angular way to do it. Use viewChild for this.

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (this.workItem.relationalData.iteration.id === this.selectedIteration.id) {
```
sanbornsen: In case there is no iteration associated with the item, this.workItem.relationalData.iteration.id will throw an error.

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
```diff
@@ -0,0 +1,81 @@
...
+      commit.setAttribute("disabled", "true");
```
sanbornsen: Better approach would be not to show the iteration in the list at all.

debloper: This is about detecting if the user has *changed* iteration. Only then enable the button and allow putting the change request to the API.

This particular condition catches:
1. WI has "iteration 1" as the associated iteration
2. User chooses "iteration 2" (button is enabled to change iteration)
3. But then the user switches back to "iteration 1"
4. **Now, because the associated and selected iterations are same - the button is disabled.**
5. Hence, it saves associating WI with an iteration it's already associated with (extra http-req).

Now, I understand your point (that don't show the current in the list iteration at all) but
1. Now you have an array-gymnastics to work with on each view update (which can regress horribly in future depending on slightest of model change)
2. As a user, I get paranoid when I can't set the old settings back in a control panel. Having the option to reset when I didn't meant to commit to a choice (and be sure no changes were applied, as the button goes disabled) is quite relieving.

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
debloper: Good catch! :D

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
debloper: Right, was thinking whether it'll be too much, but you're right.

src/app/work-item/work-item-iteration-association-modal/work-item-iteration-association-modal.component.ts
vikram-raj: When I try to reassociate the WI with the same iteration which I selected previously then I get an error cannot get property of undefined. So you need to handle that.

debloper: Good catch, fixed that interaction in last patch.

nimishamukherjee:
[test]

PR #930, opened 2017-02-10
WIP Use /api/users endpoint instead of /api/identities for fetching list of users

sanbornsen:


src/app/models/user.ts
```diff
@@ -2,7 +2,15 @@ export class User {
   attributes: {
     fullName: string;
     imageURL: string;
+    email: string;
+    bio: string;
+    provider: string;
+    url: string;
+    username: string
```
vikram-raj: Semicolon missing.

PR #932, opened 2017-02-10
Relogin functionality added

sanbornsen:
Any currently any 401 response redirects the user to the login page. 

We still need to find a better practice on UX and UI code as well.

### DIFF
--- a/src/app/auth/authentication.service.ts
+++ b/src/app/auth/authentication.service.ts
@@ -8,17 +8,17 @@ import { Broadcaster } from '../shared/broadcaster.service';
 export class AuthenticationService {
   private authToken: string = '';
 
-  constructor(private router: Router, 
-              private logger: Logger, 
+  constructor(private router: Router,
+              private logger: Logger,
               private broadcaster: Broadcaster) {
-    
+
   }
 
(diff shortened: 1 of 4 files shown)


sanbornsen:
[test]

alexeykazakov:
I wonder if calling /api/login/authorize directly instead of routing to the login page would be better approach. No need to click on login button for the user and even automated token re-issuing may work.

src/app/auth/authentication.service.ts
```diff
@@ -31,11 +31,13 @@ export class AuthenticationService {
     return false;
   }
 
-  logout() {
+  logout(redirect: Boolean = false) {
```
joshuawilson: this should be lowercase 'boolean' to be a type

sanbornsen:
@alexeykazakov I thought the same, it's just am an approach to fixing the breakdown in production. Once we have the accepted behavior we can implement it the right way. What do you say?

sanbornsen:
https://github.com/fabric8io/fabric8-planner/issues/941

alexeykazakov:
@sanbornsen, agreed.

PR #942, opened 2017-02-13
Work item description view toggle added

sanbornsen:
View toggle feature for work item description is added. In the details page, one can toggle the description view fro Markdown to HTML.

### DIFF
--- a/src/app/work-item/work-item-detail/work-item-detail.component.html
+++ b/src/app/work-item/work-item-detail/work-item-detail.component.html
@@ -211,6 +211,8 @@
         <!--Description-->
         <div class="separator col-md-12 col-sm-12 col-xs-12 detail-description-wrap">
           <div class="col-md-12 col-sm-12 col-xs-12 detail-description" *ngIf="loggedIn">
+            <button class="view-switch-btn" *ngIf="descViewType=='markdown'" (click)="showHtml(desc.innerText)">View HTML</button>
+            <button class="view-switch-btn" *ngIf="descViewType=='html'" (click)="descViewType='markdown'">View Markdown</button>
             <div class="detail-desc-div-wrap" (click)="openDescription(false, true)">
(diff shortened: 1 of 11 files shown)


src/app/work-item/work-item-detail/work-item-detail.component.html
```diff
@@ -211,7 +211,9 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
         <!--Description-->
         <div class="separator col-md-12 col-sm-12 col-xs-12 detail-description-wrap">
           <div class="col-md-12 col-sm-12 col-xs-12 detail-description" *ngIf="loggedIn">
-            <div class="detail-desc-div-wrap" (click)="openDescription(false, true)">
+            <button class="view-switch-btn" *ngIf="descViewType=='markdown'" (click)="showHtml(desc.innerText)">View HTML</button>
```
vikram-raj: Can we make this button disable until there is description text?

nimishamukherjee:
If there is no description text, and I click on it, 'undefined' is displayed.
descundefined

nimishamukherjee:
Insert a quote behaviour:
On Git:
> a
> a
> a

In work item description:
descquote

nimishamukherjee:
Functional test fails
descfuncerror

sanbornsen:
@nimishamukherjee yeah that's a script injection test. We haven't decided about the current behavior of scripts in the description. Also, the code tag is not handled from the backend.

src/app/work-item/work-item-detail/work-item-detail.component.html
sanbornsen: We can but doesn't have any use. Besides, the UX is not defined, it's just the functionality that we have implemented. Once the UX in place we'll have to rework on buttons and stuff so currently the experience is very basic and working model

Issue #871, opened 2017-02-02
Show toast notification on start iteration.


vikram-raj:
Related to: #647 

UX: https://redhat.invisionapp.com/share/FR9R8G6B9#/screens/211643014

PR #943, opened 2017-02-13
Add toast notification on start iteration.

vikram-raj:
#871

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -3,3 +3,5 @@
     <router-outlet></router-outlet>
     <alm-app-footer class="width100 pull-left"></alm-app-footer>
 </div>
+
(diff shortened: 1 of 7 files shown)


nimishamukherjee:
How long does the toast notification stay is the user does not dismiss it?

nimishamukherjee:
[test]

Essjaysee:
@nimishamukherjee Our understanding in UXD was that toast notifications were below the cut line (and not going to be in for the current scope).   Did that change?  Details about interactions for toast notifications, like how long to show it, can be found here: http://www.patternfly.org/pattern-library/communication/toast-notifications/#/api for details about the interactions.  Thanks!

catrobson:
We also should building these as a common component in the Platform, not just into the planner.

PR #946, opened 2017-02-14
POC :: suites configuration for funtional tests 

naina-verma:
This PR is related to : https://github.com/fabric8io/fabric8-planner/issues/830
Do not merge this PR it need to evaluate , for now it has SmokeTests will be adding more in to it!
We need to distribute Test cases in parts

  - Smoke
  - Sanity
  - Regression
    So that we can reduce the time of functional tests while testing PR's on Build server

### DIFF
--- a/run_functional_tests.sh
+++ b/run_functional_tests.sh
@@ -32,6 +32,8 @@ echo done.
 # Finally run protractor
 echo Running tests...
 node_modules/protractor/bin/protractor protractor.config.js
+## Run functional tests on the bases of suite genere!  ##
+#node_modules/protractor/bin/protractor protractor.config.js --suite $1 //This line has been commented for now! 
 TEST_RESULT=$?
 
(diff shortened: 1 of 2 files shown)


run_functional_tests.sh
```diff
@@ -32,6 +32,8 @@ echo done.
 # Finally run protractor
 echo Running tests...
 node_modules/protractor/bin/protractor protractor.config.js
+## Run functional tests on the bases of suite genere!  ##
+#node_modules/protractor/bin/protractor protractor.config.js --suite smokeTest //This line has been commented for now!
```
ldimaggi: I think that what makes the most sense here is to pass the test suite name to the run_functional_tests.sh as a parameter:

`
node_modules/protractor/bin/protractor protractor.config.js --suite $1`

naina-verma:
@ldimaggi , Done Changes !

PR #949, opened 2017-02-15
Added tree list component

dlabrecq:
Tree nodes are not displayed by default. To enable the toggle, the WI object must have a children array. Alternatively, we can use the components getChildren function to fetch as needed. In that scenario, a hasChildren boolean can be added to the WI object to enable the toggle.

Dragging is enabled, but can be restricted using the allowDrop function.

See below for documentation regarding available options.
https://angular2-tree.readme.io/docs/options

See below for full documentation.
https://angular2-tree.readme.io/docs

Tree list with toggle
alm-tree-list

Tree list without toggle (default)
alm-list

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "@angular/upgrade": "2.1.1",
     "@types/lodash": "4.14.41",
     "angular-in-memory-web-api": "0.1.13",
+    "angular2-tree-component": "^2.7.0",
     "bootstrap": "^3.3.7",
     "core-js": "^2.4.1",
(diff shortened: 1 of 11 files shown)


michaelkleinhenz:
**DO NOT MERGE.** We will split up that code into the parts we will have in the Planner and the parts we need to have in the shared component library.

joshuawilson:
[test]

joshuawilson:
[test]

joshuawilson:
Closing as this work was also done in #991 and merged to the ngx-widgets branch.

PR #950, opened 2017-02-15
Fixes the curious case of almEditable (on-demand) element focus

debloper:
Disabled almEditable automatically setting focus (often leading to weird interactions). With this patch, setting focus on the editable element is part of the interaction-flow implementation, and not default to almEditable.

This also fixes the existing cases affecting the changes in detail-view (title and descriptions - for general view as well as creating new WI).


nimishamukherjee:
[test]

nimishamukherjee:
[test]

PR #951, opened 2017-02-16
Fix phantom border removal

mindreeper2420:
fixes the removal of the divider between checkbox and icons when the row is selected


nimishamukherjee:
[test]

nimishamukherjee:
[test]

PR #955, opened 2017-02-16
Match Quick Add to Visual Design

mindreeper2420:
This PR is for matching the Quick Add of a Work Item to the visual design as given by UXD. The design can be found here: [Visual Design for Quick Add](https://redhat.invisionapp.com/share/DF8ZQ11R7#/196006933_A-000). Specs are given in the design document.


nimishamukherjee:
[test]

PR #991, opened 2017-02-21
Added treelist component

dlabrecq:
Tree nodes are not displayed by default. To enable the toggle, the WorkItem object must have a "children" array. Alternatively, the components getChildren function may be used to fetch children as needed. In this scenario, a "hasChildren" boolean property can be added to the WorkItem object to enable the toggle.

Note that this was tested in the master branch first, then ported to the ngx-widgets branch -- screen shots were taken from the master branch -- see PR #949.

Dragging is enabled, but can be restricted using the allowDrop function.

See below for documentation regarding available options.
https://angular2-tree.readme.io/docs/options

See below for full documentation.
https://angular2-tree.readme.io/docs

Tree list with toggle
alm-tree-list

Tree list without toggle (default)
alm-list

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "ng2-dnd": "2.0.1",
     "ng2-dropdown": "0.0.15",
     "ng2-modal": "0.0.22",
-    "ngx-widgets": "0.2.17",
+    "ngx-widgets": "0.2.18",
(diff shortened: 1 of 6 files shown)


michaelkleinhenz:
Arrrgghhhhh!

PR #1000, opened 2017-02-23
move away from the private ci registry

kbsingh:
move to the dev registry that can then be accessed from multiple clusters


joshuawilson:
[test]

PR #1010, opened 2017-02-25
chore(build): Add semantic-release

pmuir:
* Add semantic-release for continuous publishing
* Add commitizen and validate-commit-msg as helpers
* Cleanup Dockerfile.builder to reduce build times
* Don't directly use the homedir in the docker container, it can confuse npm

NOTE: I don't think I wired up the module build properly - need @joshuawilson to advise on the necessary build process


package.json
```diff
@@ -29,6 +30,7 @@
     "rimraf": "rimraf",
     "setup": "npm install && npm run validate",
     "sass-lint": "sass-lint -c .sass-lint.yml 'src/**/*.scss' -v -q",
+    "semantic-release": "semantic-release pre && npm run build:prod && cp -r .git dist && npm publish dist/ && semantic-release post",
```
joshuawilson: we want 'build' here not 'build:prod'

PR #1017, opened 2017-02-27
Fixed mock http end points and made it available for external packages using http services

sanbornsen:


joshuawilson:
This changes the global to default to true and never be set to false. So we are always in test mode?

btw, what was wrong with using observables?

sanbornsen:
@joshuawilson even before the subscriber could find the mode, the services were firing up the API  and considering it a production mode it was hitting the API URL instead of mock HTTP. Try running it into in-memory mode, you'll understand the problem.

michaelkleinhenz:
@joshuawilson mmh..I think the mode gets set by the env. The default is just the initializing. On the Observables: the mode value is not asynchronously established, why use Observables?

joshuawilson:
if you want to asynchronously update the global state then you read initial values from a config.json or remote server and you have an observer that emits those values

joshuawilson:
The use of rxjs is philosophical in nature. I prefer it. I think functional programming is a better style. That said, what you have done works and we need to keep moving. So I'm going to merge it. If something doesn't work in this, then we fix it.

PR #1022, opened 2017-02-28
fix(tests): Mock iteration add

naina-verma:
Removed a line from code which creates duplicate iterations


michaelkleinhenz:
[test]

PR #1028, opened 2017-03-01
fix(api_url): The production API URL is updated

sanbornsen:


config/webpack.prod.js
```diff
@@ -21,7 +21,7 @@ const webpackMerge = require('webpack-merge');
  * Webpack Constants
  */
 const ENV = process.env.NODE_ENV = process.env.ENV = 'production';
-const API_URL = process.env.API_URL || 'http://api.almighty.io/api/';
+const API_URL = process.env.API_URL || 'http://api.prod-preview.openshift.io/api/';
```
joshuawilson: This should be changed to 
`const FABRIC8_WIT_API_URL = process.env.FABRIC8_WIT_API_URL;`

and the URL should not be set in Prod but you should set it in dev and use the same name.

sanbornsen: I don't understand because in DEV we use localhost URL. 
So do we introduce a new ENV variable? Why would you give WIT (work item) in the variable name?

joshuawilson: @pmuir can you please comment on this.

pmuir: @sanbornsen 

1) Locally you can use either localhost or a demo server. 
2) This is not a new env var, it's a rename of API_URL
3) We renamed it because this is not the only API in the world, so we wanted to namespace the variable name. FABRIC8_WIT = Fabric8 Work Item Tracker - what we now called almighty-core

PR #1042, opened 2017-03-01
Side panel uxd

mindreeper2420:
This PR is to update the side panel UI to match the designs from UX.

As part of this work, I noticed that the kabob styling is off, but the fix affects the entire UI (not just those in the side panel). I have left that fix for a future PR.


joshuawilson:
@mindreeper2420 please squash the merge commits.

PR #1049, opened 2017-03-02
fix(router): Planner routing pattern fixed

sanbornsen:


joshuawilson:
Closing as we need to change this library not the calling application

PR #1052, opened 2017-03-02
fix(spaces) use spaceChanged

pmuir:


pmuir:
Let's not do this, there are too many race conditions that the current code is swerving around for this to be safe right now

PR #1059, opened 2017-03-03
Update WI empty state to match UXD

mindreeper2420:
The empty state for the Work Item list should match the designs as presented in the [Epics Visuals](https://redhat.invisionapp.com/share/KAAP5JDPX#/215407327_A-1000-Epics_Blank_Slate), with the difference being that this is a Work Item, and the Add actions are in the Quick Add and the '+' in the Filter Bar.

If you wish to show the message, comment out the `<alm-tree-list>` block of work-item-list.component.html


SMahil:
LGTM

PR #1087, opened 2017-03-06
Moved Patternfly styles to tree list component

dlabrecq:
Moved a couple Patternfly styles (list-view-pf, list-group & list-group-item) from the planner UI to the tree list component.

We will want to use these Patternfly styles in other places with the tree list. This means there are far less tree list styles to override in planner and/or elsewhere.


michaelkleinhenz:
[test]

PR #1091, opened 2017-03-06
Use field labels defined in work item type in WI detailed view

kwk:


src/app/work-item/work-item-detail/work-item-detail.component.html
```diff
@@ -96,7 +96,7 @@ <h2 *ngIf="loggedIn" id="wi-detail-title-click" class="detail-title truncate">{{
           </div>
         </div>
         <div class="separator col-md-12 col-sm-12 col-xs-12 detail-assignee-wrap">
-            <span class="pull-left user-label">Assignee</span>
+            <span class="pull-left user-label">{{workItem.relationalData.wiType?.attributes?.fields['system.assignees']?.label}}</span>
```
kwk: This will actually show up as the plural form `"Assignees"`. Shall we fix it in core to only show as a singular `"Assingee"`?

michaelkleinhenz:
[test]

PR #1095, opened 2017-03-06
fix(board): Fix filter bar style.

vikram-raj:


mindreeper2420:
@vikram-raj @michaelkleinhenz would it be possible to add a screenshot of the before and after to these (or the Issues)? I'll do the same when I file a PR :-)

PR #1099, opened 2017-03-06
Sass2

pmuir:


pmuir:
Please wait and see if tests will pass.

michaelkleinhenz:
Looks like there is something fishy with the loaders..

pmuir:
Yep, fixing.

config/webpack.common.js
```diff
@@ -82,36 +88,59 @@ module.exports = {
...
+        loader: extractCSS.extract({
+          fallback: "style-loader",
+          use: "css-loader?sourceMap&context=/"
+        })
       },
+
       {
         test: /\.scss$/,
-        use: [
+        loaders: [
```
joshuawilson: Why change this to "loaders", under webpack 2.2 it should be "use". The allow "loaders" for backwards compatibility but it is no longer the standard.

joshuawilson:
@michaelkleinhenz if you approve it, I can merge it without adding a Merge commit. I confirmed there are not merge conflicts.

PR #1152, opened 2017-03-13
Adding mock for area Label

naina-verma:
Adding mock for area Label


naina-verma:
@nimishamukherjee , i need this change for automation specifically 'Areas' , can we have another PR in which we can add other's ?
Currently , i am focused on Area's only

PR #1161, opened 2017-03-14
fix(planner-height): fixes planner work item list height

mitchharpur:
This is a workaround that adds height-100 to the list too ensure it fits
the container


sanbornsen:
There is already a PR to fix this - https://github.com/fabric8io/fabric8-planner/pull/1166

Hence closing this one.

PR #1167, opened 2017-03-14
feat(routing): Lazy loading to board view added

sanbornsen:
Go ahead and merge it !!


sanbornsen:
Not now

package.json
```diff
@@ -129,8 +129,8 @@
     "@types/selenium-webdriver": "2.53.41",
     "@types/zone.js": "0.0.27",
     "angular-in-memory-web-api": "0.3.0",
-    "angular2-load-children-loader": "0.1.3",
-    "angular2-router-loader": "0.3.5",
+    "angular2-load-children-loader": "^0.1.3",
```
joshuawilson: Please do not add any auto-updates to the dependencies. Now that we are running a shared system, it will be lighter weight if they all use the same. And it will introduce less bugs.

sanbornsen: Yeah true, make sense. I'll update that.

michaelkleinhenz:
[test]

PR #1201, opened 2017-03-15
Added iteration definition to support iterations tests that rely on m…

ldimaggi:
…ocking


naina-verma:
screenshot from 2017-03-16 11-19-48

PR #1240, opened 2017-03-21
feat(listview): integrate order of execution api in work item list view.

vikram-raj:
This PR contain Order of Execution only by drag and drop.


vikram-raj:
Merge this PR after https://github.com/fabric8-ui/ngx-widgets/pull/51 PR.

PR #1252, opened 2017-03-22
fix(treelist): Refactored treelist to better mimic dnd mock-ups

dlabrecq:
Note that this depends on this ngx-widgets PR:
https://github.com/fabric8-ui/ngx-widgets/pull/51

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.html
+++ b/src/app/work-item/work-item-list/work-item-list.component.html
@@ -20,17 +20,17 @@
                 [loadTemplate]="treeListLoadTemplate"
                 [nodes]="workItems"
                 [options]="treeListOptions"
+                [showDropSlot]="false"
                 [showExpander]="false"
-                (onMoveNode)="onMoveNode($event)"
-                [showDropSlot]="true">
+                (onMoveNode)="onMoveNode($event)">
               <template #treeListLoadTemplate let-node="node" let-index="index">
                 <span>Loading...</span>
               </template>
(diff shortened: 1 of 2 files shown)


dlabrecq:
FYI @vikram-raj, made some final style tweaks after testing ngx-widgets in the planner ui.

michaelkleinhenz:
Should we wait to merge this until fabric8-ui/ngx-widgets#51 is merged? @dlabrecq

dlabrecq:
Yes. We need to merge ngx-widgets, first. @joshuawilson ?

michaelkleinhenz:
Interesting is that the check has succeeded here. It should not. We had that on another PR from @vikram-raj  too and merged it, so current master requires the new ngx-widgets version and are broken right now.
@ldimaggi can you take a look why the tests here did not fail?

michaelkleinhenz:
[test]

dlabrecq:
FYI, the planner UI only requires the latest ngx-widgets so it can achieve the desired layout shown in the mock-ups. I refactored tree-list to add some new attributes to achieve this. However, it won't generate an error if they are not used.

I don't expect the tests would fail. Instead, the tree list simply won't visually appear the way we expect it to in the planner UI. For example, you may see extra white space between rows and drag and drop may not function properly.

dlabrecq:
Rebased

dlabrecq:
@michaelkleinhenz Can this be merged? fabric8-ui/ngx-widgets#51 has been merged and I've rebased.

PR #1265, opened 2017-03-23
fix(version): update ngx-widgets version.

vikram-raj:


package.json
```diff
@@ -117,7 +117,7 @@
     "ngx-fabric8-wit": "6.15.2",
     "ngx-login-client": "0.6.5",
     "ngx-modal": "0.0.29",
-    "ngx-widgets": "0.9.6",
+    "ngx-widgets": "0.10.0",
```
dlabrecq: This will need to be 0.10.1 to get the latest tree-list changes. 

FYI, I've actually updated PR #1252 to use the latest ngx-widgets package

vikram-raj: @dlabrecq ok will update this.

PR #1337, opened 2017-03-27
test(board-spec-remove): remove board spec temporarily

naina-verma:
Due to this bug https://github.com/fabric8io/fabric8-planner/issues/1296


naina-verma:
@ldimaggi can you plz look into it!

Issue #1246, opened 2017-03-22
Make big plus symbol on empty WI list clickable


kwk:
I would love to see that this big plus symbol that is JUMPING IN MY FACE is clickable. But currently it isn't.

![screenshot from 2017-03-22 14-17-04](https://cloud.githubusercontent.com/assets/193408/24202158/9392050c-0f12-11e7-8ff8-6d37422343c2.png)

@Mgranfie this would increase the usability, no?

mindreeper2420:
@kwk can you see if it is just as bad when Epics are empty? We are using the same empty state message for that as well. I don't notice those things, since I did them :)

kwk:
> @kwk can you see if it is just as bad when Epics are empty?

I don't get what you're saying.

Mgranfie:
This is good feedback Konrad... will look into this!

mindreeper2420:
@kwk ignore my comment - doesn't apply 😄

maxandersen:
+1 on this one - its the biggest target on the screen but it doesn't do anything. +1 on making it actually "Add" or remove it.

debloper:
Fixed with #1685

PR #1349, opened 2017-03-28
fix(list view): Make the plus symbol clickable ( #1246).

nimishamukherjee:
Make the plus symbol click-able and display the work item types page. the The plus symbol appears in the work item list view when there are no work items.

### DIFF
--- a/src/app/work-item/work-item-list/work-item-list.component.html
+++ b/src/app/work-item/work-item-list/work-item-list.component.html
@@ -1,3 +1,4 @@
+<detail-add-type-selector #typeSelectPanel></detail-add-type-selector>
 <div class="container-fluid work-item-page">
   <main class="row">
     <aside>
@@ -63,6 +64,9 @@
                     <div class="card-pf-body">
                       <h4>There are no Work Items for your selected criteria.</h4>
                       <p>If you'd like to add a Work Item, you can do so using the Quick Add or the <i class="pficon pficon-add-circle-o"></i> in the bar above.</p>
+                       <div class="blank-slate-pf-main-action">
+                         <button class="btn btn-primary btn-lg" (click)="typeSelectPanel.openPanel()" > Create Work Item </button>
+                        </div>
                     </div>
                   </div>
                 </div>
--- a/src/app/work-item/work-item-list/planner-list.module.ts
+++ b/src/app/work-item/work-item-list/planner-list.module.ts
@@ -28,6 +28,7 @@ import { IterationModule } from '../../iteration/iteration.module';
 import { PlannerListRoutingModule } from './planner-list-routing.module';
 import { SidepanelModule } from '../../side-panel/side-panel.module';
 import { ToolbarPanelModule } from '../../toolbar-panel/toolbar-panel.module';
+import { WorkItemDetailAddTypeSelectorModule } from '../../work-item/work-item-detail-add-type-selector/work-item-detail-add-type-selector.module';
 import { WorkItemDetailModule } from '../work-item-detail/work-item-detail.module';
 import { WorkItemListComponent } from './work-item-list.component';
 import { WorkItemListEntryComponent } from './work-item-list-entry/work-item-list-entry.component';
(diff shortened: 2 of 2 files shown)


mindreeper2420:
Hi @nimishamukherjee 

In order to keep this PR in-line with the [PatternFly component](http://www.patternfly.org/pattern-library/communication/empty-state/), I would suggest removing the clickability from the '+' and adding a primary button (per the component designs) and have that perform the action to open the Work Item type page for adding a new Work Item. We can revisit this down the line (per our continuous iteration model) to determine what to do with other areas with empty states, in order to maintain consistency. UX will continue to evaluate how this is utilized as we move forward.

mindreeper2420:
Also of note, the gray background should not be present - that is being pulled in from the Spaces template.

nimishamukherjee:
@mindreeper2420 to make sure I understand what you are saying:
Based on http://www.patternfly.org/pattern-library/communication/empty-state/#/overview
I add the "Main Action" button, maybe with title "Add", clicking on this will initiate the detailed add flow.

mindreeper2420:
@nimishamukherjee yes - I would make the text in the button read 'Create Work Item', so that it matches the Quick Add text.

Mgranfie:
@mindreeper will this put focus on the quick add or the full details pane
will open?

michaelkleinhenz:
@Mgranfie @mindreeper2420 can you mark the PR as "changes requested" if there need things to be worked on. It is hard to see from discussions if there is actually objections present. thx!

src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -52,7 +53,7 @@
                 <div class="col-sm-offset-2 col-md-offset-3 col-sm-8 col-md-6">
                   <div class="card-pf card-pf-aggregate-status">
                     <div class="card-pf-title">
-                      <h2 class="pficon pficon-add-circle-o no-wi"></h2>
+                      <h2 class="pficon pficon-add-circle-o no-wi with-cursor-pointer" (click)="typeSelectPanel.openPanel()"></h2>
```
mindreeper2420: this should not be clickable. per the PatternFly Empty State component, there should be a clickable 'Create Work Item' primary button below all of the text.

src/app/work-item/work-item-list/work-item-list.component.html
```diff
@@ -52,7 +53,7 @@
                 <div class="col-sm-offset-2 col-md-offset-3 col-sm-8 col-md-6">
                   <div class="card-pf card-pf-aggregate-status">
```
mindreeper2420: please remove the gray background that is present - according to the inspector, it is coming from the Spaces component scss file.

nimishamukherjee:
@Preeticp, @mindreeper2420 once we add the blue Create Work Item button, we would have three options to create a work item.

createwi

The second line - "If you'd like to add a Work Item, you can do so using the Quick Add or the + in the bar above."
needs to be modified.

Concerns:
- The user does not understand what a quick add is
- The + icon appears in multiple places

Preeticp:
A few suggestions/questions:

1. Do we need the black plus icon right in the middle of the page?
2. Maybe, the blue 'Create work item' button could replace the text in black, below the plus icon, 'Create a Work Item', making it obvious.
3. We could simply add 1 line saying, 'you could also use the Quick add (or the icon representing it) to create a Work Item. 
4. The Quick Add should have a unique icon which could be consistently used to represent smart/quick add actions throughout the UI.

.

nimishamukherjee:
@michaelkleinhenz @mindreeper2420 can I close this PR. The original issue was about making the plus icon click-able and this involves a discussion with the PatternFly team.
Adding the Create Work Item button may complicate this issue further as the instruction text needs to be updated.
Your thoughts?

mindreeper2420:
@nimishamukherjee @michaelkleinhenz I would close this and leave it as is, for the time being. All of the empty states need to be adjusted throughout the product.

nimishamukherjee:
Closing this PR.

PR #1351, opened 2017-03-28
More work on observable with trying to fix broken pieces in places

sanbornsen:


src/app/app-routing.module.ts
```diff
@@ -11,6 +11,10 @@ const routes: Routes = [
     pathMatch: 'full'
   },
   {
+    path: 'list',
```
sanbornsen: This is just for testing the lazy loading of list page. You can ignore this. Not effecting anything.

PR #1360, opened 2017-03-29
Functional tests running with chrome or/and optionaly using firefox

hectorj2f:
I commented the code for firefox in case you wanna try it again in the future. @ldimaggi


Dockerfile.builder
```diff
@@ -16,31 +13,47 @@ RUN set -ex \
...
-RUN yum install -y bzip2 fontconfig java-1.8.0-openjdk nmap-ncat psmisc git \
-  && curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.xz" \
-  && curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/SHASUMS256.txt.asc" \
-  && gpg --batch --decrypt --output SHASUMS256.txt SHASUMS256.txt.asc \
-  && grep " node-v$NODE_VERSION-linux-x64.tar.xz\$" SHASUMS256.txt | sha256sum -c - \
-  && tar -xJf "node-v$NODE_VERSION-linux-x64.tar.xz" -C /usr/local --strip-components=1 \
-  && rm "node-v$NODE_VERSION-linux-x64.tar.xz" SHASUMS256.txt.asc SHASUMS256.txt \
-  && ln -s /usr/local/bin/node /usr/local/bin/nodejs \
-  && yum clean all
+USER root
```
ppitonak: I believe that you don't need this line.

Dockerfile.builder
```diff
@@ -16,31 +13,47 @@ RUN set -ex \
...
+RUN yum -y update && \
+    yum install -y bzip2 fontconfig tar java-1.8.0-openjdk nmap-ncat psmisc gtk3 git \
+      python-setuptools xorg-x11-xauth wget unzip which xorg-x11-server-Xvfb \
+      xfonts-100dpi xorg-x11-fonts-75dpi xfonts-scalable xfonts-cyrillic \
+      ipa-gothic-fonts xorg-x11-utils xorg-x11-fonts-Type1 xorg-x11-fonts-misc \
+      GConf2 wget libXfont wget && \
+      yum -y clean all
+RUN yum install -y firefox
```
ppitonak: Why not to install Firefox together with other packages? This way you will pollute Yum cache after you cleaned it.

Dockerfile.builder
```diff
@@ -16,31 +13,47 @@ RUN set -ex \
...
+
+RUN wget "http://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.gz" \
+  && tar -xvf "node-v$NODE_VERSION-linux-x64.tar.gz" -C /usr/local --strip-components=1 \
+  && rm "node-v$NODE_VERSION-linux-x64.tar.gz" \
+  && ln -s /usr/local/bin/node /usr/local/bin/nodejs
+
+RUN  wget https://github.com/mozilla/geckodriver/releases/download/v0.14.0/geckodriver-v0.14.0-linux64.tar.gz && \
```
ppitonak: I suggest to extract Gecko driver's version in similar way as Node.js version on line 22.

Dockerfile.builder
```diff
@@ -16,31 +13,47 @@ RUN set -ex \
...
 ENV FABRIC8_USER_NAME=fabric8
 
 RUN useradd --user-group --create-home --shell /bin/false ${FABRIC8_USER_NAME}
 
 ENV HOME=/home/${FABRIC8_USER_NAME}
-
 ENV WORKSPACE=$HOME/fabric8-planner
 RUN mkdir $WORKSPACE
 
+RUN chmod u+s /usr/bin/Xvfb && chown fabric8 /usr/bin/Xvfb
```
ppitonak: 1. What is this line good for?
2. Is shouldn't be 'fabric8' string replaced with `${FABRIC8_USER_NAME}`?

Dockerfile.builder
```diff
@@ -49,4 +62,9 @@ WORKDIR $WORKSPACE/
 
 VOLUME /dist
 
-ENTRYPOINT ["/bin/bash"]
+COPY docker-entrypoint.sh /home/fabric8/fabric8-planner/
```
ppitonak: 1. If I understand it correctly, you don't need to copy `docker-entrypoint.sh` because all files are already copied (see line 57)
2. `/home/fabric8/fabric8-planner` should be replaced with `$WORKSPACE`

Dockerfile.builder
```diff
@@ -49,4 +62,9 @@ WORKDIR $WORKSPACE/
...
+
+# Open ports
+EXPOSE 4444
```
ppitonak: Why?

hectorj2f: it isn't necessary

Dockerfile.builder
hectorj2f: 1 is enough :)

Dockerfile.builder
hectorj2f: perhaps we won't use firefox, WIP

Dockerfile.builder
hectorj2f: indeed I don't need to be root user.

Dockerfile.builder
```diff
@@ -16,31 +13,51 @@ RUN set -ex \
     B9AE9905FFD7803F25714661B63B535A4C206CA9 \
     C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
   ; do \
-    gpg --import "/gpg/${key}.gpg" ; \
```
hectorj2f: @naina-verma any reason why this changed in your branch ? I just took what you had in your branch.

Dockerfile.builder
hectorj2f: 1. We might not need it anymore
2. ...it could if I need 1.

Dockerfile.builder
ldimaggi: +1 Firefox has problems - many people are complaining about version mismatches - let's focus on Chrome first!

Dockerfile.builder
naina-verma: @hector , this may be typo i didn't made it intentionally !

hectorj2f:
I am now happy with the code ;)

hectorj2f:
@ldimaggi should we merge this PR?

PR #1392, opened 2017-03-31
fix(https): Replace all the env based url to WIT_API_URL 

sanbornsen:


src/app/work-item/work-item-detail/work-item-link/search-data.ts
```diff
@@ -1,37 +0,0 @@
-import { Http, Response } from '@angular/http';
```
sanbornsen: Yes. No need of that file anymore.

PR #1414, opened 2017-04-01
fix: use navigate with query params

pmuir:


pmuir:
Merging as no one is around to review.

PR #1506, opened 2017-04-05
fix(iterationpanel): repaired the iteration panel.

michaelkleinhenz:
Several fixes:
- selected iteration (or backlog) is now highlighted.
- backlog item count is not fake anymore.
- backlog item numbers and iteration item numbers get properly updated on work item actions.
- cursor is now a pointer over the iteration entries.
- the whole cell is clickable, not only the title.


joshuawilson:
[test]

PR #1509, opened 2017-04-05
fix(workitemlist): disabled checkboxes.

michaelkleinhenz:
Removed the checkboxes on the work item list. We do now allow multiple selections yet anyway.

### DIFF
--- a/src/app/toolbar-panel/toolbar-panel.component.html
+++ b/src/app/toolbar-panel/toolbar-panel.component.html
@@ -31,6 +31,7 @@
             </li>
           </ul>
         </div>
+        <!--
         <span *ngIf="context==='listview'" class="dropdown primary-action" dropdown>
           <button class="btn btn-default" id="wi_filter_dropdown" type="button" dropdownToggle>
             Move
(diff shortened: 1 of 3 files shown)


maxandersen:
now -> not, right ? :)

mindreeper2420:
If you remove the check boxes, then what happens to the 'Move' button in the toolbar?

michaelkleinhenz:
I'd say lets keep the move dropdown and change the behaviour that it touches the currently selected entry. After all, we still "select" an entry when clicking on it. @mindreeper2420 ok with that?

nimishamukherjee:
For the touch/select we have the drag behaviour for re-ordering.

mindreeper2420:
@michaelkleinhenz that was not the intention of the original design, as opening a work item was not selecting it for movement. Please reach out to @catrobson for clarification, as she completed the latest designs for the toolbar.

catrobson:
Design intent:
Checkboxes select the row, but do not open the work item.  
Checkboxes only show on hover, but once check remain on screen so you can see all the items you have checked.
Example:  https://redhat.invisionapp.com/share/YVA4PGYA3#/225834565_Multi-Select

I think there are two implementation options:
1. Remove the checkboxes and the move menu completely from the system if we cannot support multi-select in a timely manner
2. Keep the checkboxes and keep the move menu as designed

michaelkleinhenz:
I changed the PR to option #1. I think this makes the most sense.

Issue #1501, opened 2017-04-05
On https://prod-preview.openshift.io - creating a new iteration displays the root iteration as "null


ldimaggi:
![screenshot from 2017-04-05 14-45-09](https://cloud.githubusercontent.com/assets/642621/24721474/7c5d56d2-1a0e-11e7-8bbc-be88b6797814.png)

vikram-raj:
Fixed in #1520

PR #1520, opened 2017-04-06
fix(iteration): issue in nested iteration

vikram-raj:
- Fix issue - #1501 

- match search parent iteration field UX with [wireframe](https://redhat.invisionapp.com/share/YPB0J2NGB#/screens/225825534).

- Prefilled the parent iteration field if user select create child option from kebob menu of existing iteration.


vikram-raj:
Work in progress on this PR.

PR #1534, opened 2017-04-06
fix(iteration): show date picker in create interation modal (fix #1341)

### LINKED ISSUE #1341: In http://prod-preview.openshift.io, start/end dates cannot be specified when an iteration is created - but default values are assigned
See attached screenshots - the user is not given the option of defining start and/or end dates when creating a new iteration. Default values are assigned and the user must edit the iteration to change these dates.

screenshot from 2017-03-27 09-26-47
screenshot from 2017-03-27 09-27-31

vikram-raj:
Previously we didn't allow the user to choose iteration date interval while creating the new iteration. Now this PR will allow the user to select the iteration date interval while creating the new iteration.

### DIFF
--- a/src/app/iteration/iteration-modal/iteration-modal.component.html
+++ b/src/app/iteration/iteration-modal/iteration-modal.component.html
@@ -47,7 +47,7 @@
         </div>
       </div>
 
-      <div *ngIf="modalType=='update' || modalType=='start' || modalType=='createChild'"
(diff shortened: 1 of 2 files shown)


michaelkleinhenz:
The date pickers should be empty on new iterations, not filled with a default date. Is that the case with this change? If no, can you add that? Thanks!

vikram-raj:
@michaelkleinhenz No, now date picker filled with a current date. and yes, I am adding that.

nimishamukherjee:
Current date/default does not show up when creating a new iteration:
dates

PR #1559, opened 2017-04-10
fix(release): set release debug to false so fabric8 release pipelines…

rawlingsj:
… work with NPM registry


rawlingsj:
@joshuawilson this should mean the downstream projects will be updated with the new release version

joshuawilson:
[test]

PR #1561, opened 2017-04-10
Fix iterations, Part 2

michaelkleinhenz:
Fixes several more issues in the iterations panel:
- when the iteration on a work item changes, the counts on the iteration panel are updated.
- the root iteration is no longer displayed in the iteration list.
- several severe bug fixes in the mock data service(s).
- the date display is not visible when no dates are set (no lone "-" display on the current iteration in that case).

This also fixes some other things:
- the list is now updated correctly when selecting a new iteration (only in list view, in board view, this is still broken).
- fixes some other mock related problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "@angular/platform-browser-dynamic": "4.0.1",
     "@angular/platform-server": "4.0.1",
     "@angular/router": "4.0.1",
-    "angular-tree-component": "3.2.4",
+    "angular-tree-component": "3.2.3",
     "core-js": "2.4.1",
     "lodash": "4.17.4",
     "moment": "2.18.1",
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -79,10 +79,11 @@ if (process.env.ENV == 'inmemory') {
     serviceImports,
     SpacesService,
     ssoApiUrlProvider,
+    MockHttp,
     DummySpace,
     {
       provide: HttpServiceLGC,
-      useClass: MockHttp
+      useExisting: MockHttp
     }
   ];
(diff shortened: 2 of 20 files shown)


package.json
```diff
@@ -106,7 +106,7 @@
     "@angular/platform-browser-dynamic": "4.0.1",
     "@angular/platform-server": "4.0.1",
     "@angular/router": "4.0.1",
-    "angular-tree-component": "3.2.4",
+    "angular-tree-component": "3.2.3",
```
nimishamukherjee: Platform is using "angular-tree-component": "3.2.4"

michaelkleinhenz: Ah, yes. Have to check that again. 3.2.4 was broken.

michaelkleinhenz:
Weird, I checked the integrated variant, and it was working. Will check.

michaelkleinhenz:
@nimishamukherjee ok, the core reports wrong numbers currently. @pranavgore09 already has a PR in place to fix them. This PR is displaying the "correct" numbers reported by the core. So all ok here.

michaelkleinhenz:
@nimishamukherjee on the tree component: the problem there is still not fixed. We either break the tests or prod with the new tree component version.

fabric8cd:
@michaelkleinhenz snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1561.int.rdu2c.fabric8.io

fabric8cd:
@michaelkleinhenz snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1561.int.rdu2c.fabric8.io

PR #1576, opened 2017-04-11
fix(version): update ngx-login-client to 0.6.21

fabric8cd:


fabric8cd:
@fabric8cd snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1576.int.rdu2c.fabric8.io

PR #1588, opened 2017-04-11
fix(version): update ngx-widgets to 0.13.4

fabric8cd:


fabric8cd:
@fabric8cd snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1588.int.rdu2c.fabric8.io

PR #1605, opened 2017-04-12
fix(detail page): Display saving text during work item creation.

nimishamukherjee:
fix(detail page): Display saving text during work item creation.


nimishamukherjee:
@mindreeper2420 , @michaelkleinhenz let me know if this PR helps better the user experience.
As there is a delay on prod-preview when saving a work item, we could show a Saving work item message.

mindreeper2420:
I'll take a look at this tonight and let you know.

src/app/work-item/work-item-detail/work-item-detail.component.html
```diff
@@ -1,4 +1,7 @@
 <div [@slideInOut]="panelState" class="detail-panel">
+  <h3 *ngIf="saving">
+    <p><span class="spinner spinner-xs spinner-inline"></span>Saving work item.</p>
```
mindreeper2420: change the spinner to `spinner-sm` and the text to 'Saving Work Item...'

mindreeper2420:
I like this 😄  I requested a change to the text and to make the spinner larger - that way it will be more noticeable. I'd also suggest that the message be moved to be in-line with the Work Item type icon (where the Title would eventually reside). This will prevent a jarring effect of pushing the items down when the message pops up.

PR #1622, opened 2017-04-13
fix(dynamicform): use generic dropdown for dynamic forms.

michaelkleinhenz:
With this PR, we are using the typeahread dropdown also in the dynamic form area, further aligning the UI to a common behavior and look.


fabric8cd:
@michaelkleinhenz snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1622-fabric8-planner.int.rdu2c.fabric8.io

PR #1645, opened 2017-04-18
fix(buttons): update to use btn-primary

mindreeper2420:
update buttons with the checkmark (and those dictated by designs) to be styled as primary buttons


mindreeper2420:
I'll fix the failing tests and update.

nimishamukherjee:
@mindreeper2420 there are no failing tests.
ci.centos.org PR build — Build finished. 
The snapshot was not generated for some reason.

PR #1720, opened 2017-04-21
fix(router): Sticking to iteration filter on view change [only works …

sanbornsen:
…in planner stand alone]


fabric8cd:
@sanbornsen snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1720-fabric8-planner.int.rdu2c.fabric8.io

PR #1721, opened 2017-04-21
fix(board view): Display newly added work item, added via the detailed view, in the board view. Fixes #1496.

### LINKED ISSUE #1496: Creating a new work item board view doesn't show up
http://www.screencast.com/t/rMBJ3lChHqr

nimishamukherjee:
Display newly added work item, added via the detailed view, in the board view. Fixes #1496.

### DIFF
--- a/src/app/shared/filter.service.ts
+++ b/src/app/shared/filter.service.ts
@@ -10,8 +10,6 @@ import { HttpService } from './http-service';
 
 import { FilterModel } from '../models/filter.model';
 
-import { WorkItem } from '../models/work-item';
-
 @Injectable()
 export class FilterService {
   public filters: FilterModel[] = [];
@@ -90,7 +88,7 @@ export class FilterService {
    * @param WorkItem - workItem
    * @returns Boolean
    */
-  doesMatchCurrentFilter(workItem: WorkItem): Boolean {
+  doesMatchCurrentFilter(workItem): Boolean {
     //If filters have been applied
     if (this.activeFilters.length) {
       let matchFilterCount = 0;
--- a/src/app/shared/filter.service.ts
+++ b/src/app/shared/filter.service.ts
@@ -10,6 +10,8 @@ import { HttpService } from './http-service';
 
 import { FilterModel } from '../models/filter.model';
 
+import { WorkItem } from '../models/work-item';
+
 @Injectable()
 export class FilterService {
   public filters: FilterModel[] = [];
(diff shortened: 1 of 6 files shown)


src/app/work-item/work-item-board/work-item-board.component.ts
```diff
@@ -90,6 +90,11 @@ export class WorkItemBoardComponent implements OnInit, OnDestroy {
     private areaService: AreaService,
     private filterService: FilterService,
     private route: ActivatedRoute) {
+      router.events.subscribe(event => {
```
sanbornsen: * I believe, every time you open the details page this will trigger even if you are not adding a new item. In fact on every navigation the list will re load with this chunk of code.  
* Also it's not being unsubscribed anywhere, so that could cause a event leakage.

sanbornsen: I strongly feel there is a better way to do this, let me figure out.

nimishamukherjee: The challenge is the detail view is a router outlet so we cannot bind any listener.
If you see work-item-board.component.html > <alm-work-item-quick-add
(workItemCreate)="onCreateWorkItem($event)" takes care of updating the lane.

sanbornsen: I understand, but still no playing with routers anywhere, other than filtering and querying. Instead, we can use workitemService. We can keep a BehaviorSubject in work item service and any component like board can watch over it for update or add item.

sanbornsen: Also we need this feature matching with the current filter to the page. Right?

fabric8cd:
@nimishamukherjee snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1721-fabric8-planner.int.rdu2c.fabric8.io

nimishamukherjee:
[test]

fabric8cd:
@nimishamukherjee snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1721-fabric8-planner.int.rdu2c.fabric8.io

src/app/work-item/work-item-board/work-item-board.component.ts
```diff
@@ -16,7 +16,7 @@ import {
   ViewEncapsulation
 } from '@angular/core';
 import { Response } from '@angular/http';
-import { Router, ActivatedRoute } from '@angular/router';
+import { NavigationEnd, Router, ActivatedRoute } from '@angular/router';
```
sanbornsen: **NavigationEnd** should not be needed anymore.

src/app/work-item/work-item.service.ts
```diff
@@ -687,6 +692,18 @@ export class WorkItemService {
   }
 
   /**
+   * Usage: This method emit a new work item created event
+   * The list view and board view listens to this event
+   * and updates the view based on applied filters.
+   */
+
+  emitAddWI(workItem: WorkItem) {
+    if(this.filterService.doesMatchCurrentFilter(workItem)){
```
sanbornsen: This is okay, but think, now work item service has a dependency on filter service. We can have it in the component. No matter what item it is, it should be emitted and then from the component we check if the item matches current filter or not.
Also, we can use the same observer in other places where filter doesn't matter.

src/app/shared/filter.service.ts
```diff
@@ -10,6 +10,8 @@ import { HttpService } from './http-service';
 
 import { FilterModel } from '../models/filter.model';
 
+import { WorkItem } from '../models/work-item';
```
sanbornsen: This is a direct dependency to WorkItem module, we don't need the type here, do we?

src/app/work-item/work-item.service.spec.ts
```diff
@@ -234,10 +237,11 @@ describe('Work Item Service - ', () => {
   });
 
   beforeEach(inject(
-    [WorkItemService, MockBackend, IterationService, AreaService],
-    (service: WorkItemService, mock: MockBackend, iService: IterationService, aService: AreaService) => {
+    [WorkItemService, MockBackend, FilterService, IterationService, AreaService],
```
sanbornsen: Now, do we need **FilterService** in the spec anymore?

fabric8cd:
@nimishamukherjee snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1721-fabric8-planner.int.rdu2c.fabric8.io

fabric8cd:
@nimishamukherjee snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1721-fabric8-planner.int.rdu2c.fabric8.io

PR #1724, opened 2017-04-21
Expand iteration section on selected any iteration from that section

sanbornsen:


fabric8cd:
@sanbornsen snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1724-fabric8-planner.int.rdu2c.fabric8.io

sanbornsen:
I see, that behavior is hard to achieve. Technically the board and the list are complete separate modules. So the wrapper of sidebar and toolbar around the view looks like static and only the view is switching between the board and the list, but actually, they are getting re-rendered too. When you toggle the view, it's a new sidebar component and toolbar component rendered which has no idea of the previous state. We are maintaining the states using URL's query parameters now. 
@joshuawilson can enlight more on that.

Issue #1717, opened 2017-04-21
Regression: Iteration Panel Click Points


mindreeper2420:
In the current version of the product, a user has to specifically click on the Iteration title in order to select that iteration. In previous versions, a user could click anywhere in the Iteration area to select that specific Iteration.

![iterationpanel](https://cloud.githubusercontent.com/assets/4032718/25258414/42484a28-260d-11e7-9f70-da4f702ac071.png)

When the Iteration is active, it is highlighted as such:
<img width="490" alt="screen shot 2017-04-20 at 9 07 01 pm" src="https://cloud.githubusercontent.com/assets/4032718/25258424/5d6638a6-260d-11e7-8134-cb300984e09f.png">

naina-verma:
Verified !

PR #1766, opened 2017-04-25
fix(lists): entry list items are now full clickable to select.

### LINKED ISSUE #1764: Remove link type from iteration panel and work item names
In the iteration panel, iteration names should not be blue links and the entire iteration row should be selectable.  Also captured in this issue:  https://github.com/fabric8io/fabric8-planner/issues/1717

Also for individual work items in the list view and board view, the names should not be links, but rather just black text.  Selecting the whole row or the whole card should bring up the work item detail panel.

This PR caused this change in behavior: https://github.com/fabric8io/fabric8-planner/pull/1688

michaelkleinhenz:
Fixes #1764 and #1717

### DIFF
--- a/src/app/iteration/iteration.component.html
+++ b/src/app/iteration/iteration.component.html
@@ -11,17 +11,16 @@
         </h3>
       </div>
       <ul class="iteration-expanded-div" [collapse]="isCollapsedCurrentIteration">
-        <li class="active" *ngFor="let iteration of currentIterations" [routerLinkActive]="'selected-iteration'">
+        <li class="active" *ngFor="let iteration of currentIterations" [routerLinkActive]="'selected-iteration'" 
+                [routerLink]="[]"
+                [queryParams]="{iteration: iteration.attributes.resolved_parent_path + '/' + iteration.attributes.name}">
           <div class="iteration-subheader flex-container">
             <div class="flex-grow-1 truncate">
-              <a [routerLink]="[]"
-                [queryParams]="{iteration: iteration.attributes.resolved_parent_path + '/' + iteration.attributes.name}" >
                 <strong class="iteration-name truncate"
                   tooltip="{{iteration.attributes.name}}"
                   placement="top">
                   {{iteration.attributes.name}}
                 </strong>
-              </a>
               <div class="parent-iteration">
                 <span>
                   {{iteration.attributes.resolved_parent_path + '/' + iteration.attributes.name}}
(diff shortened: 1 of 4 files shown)


fabric8cd:
@michaelkleinhenz snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1766-fabric8-planner.int.rdu2c.fabric8.io

src/app/iteration/iteration.component.html
```diff
@@ -11,17 +11,16 @@
         </h3>
       </div>
       <ul class="iteration-expanded-div" [collapse]="isCollapsedCurrentIteration">
-        <li class="active" *ngFor="let iteration of currentIterations" [routerLinkActive]="'selected-iteration'">
+        <li class="active" *ngFor="let iteration of currentIterations" [routerLinkActive]="'selected-iteration'"
```
sanbornsen: So, problem with this code is, you can't really right click on that section and copy the URL ot open it in a new tab. So we need an anchor tag for this.

sanbornsen: So we had it almost like this previously, and that's the reason I changed them to anchor tag. It works fine, so if we think it is okay for the summit, then let's go ahead with this.

michaelkleinhenz: Uhh..We should *not* support open in new tabs anyway at this point. This causes all sorts of problems, lost updates are the least. So I think explicitly disabling that is actually a good thing.

michaelkleinhenz:
Something more general. @sanbornsen the solution with the attributes is really slick and cool. I was first a bit confused and looked how the linking works but as soon as I discovered it is just that two attributes that add routing capabilities to an element I was like "oh, and that's it? that's cool!" :-) Good Work!

fabric8cd:
@michaelkleinhenz snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1766-fabric8-planner.int.rdu2c.fabric8.io

fabric8cd:
@michaelkleinhenz snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1766-fabric8-planner.int.rdu2c.fabric8.io

src/app/iteration/iteration.component.html
sanbornsen: If that is the case then all good to go. 👍

sanbornsen:
@michaelkleinhenz thanks. Yeah it's really cool.

PR #1778, opened 2017-04-26
fix(filter): Filter UX issue fixed by adding loader

sanbornsen:


fabric8cd:
@sanbornsen snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1778-fabric8-planner.int.rdu2c.fabric8.io

fabric8cd:
@sanbornsen snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1778-fabric8-planner.int.rdu2c.fabric8.io

catrobson:
Can we use full name instead of username?  See my full name in top right, vs. the username that is showing up in the filter dropdown.  We use full dropdown in the planner detail item as well I believe.

image

sanbornsen:
@catrobson the values in filters are used to construct the URL as well. Username is unique but names are not. Imagine having two users with same name.

catrobson:
Ok, let's leave this alone for summit then.  In the future, I don't think that we should tightly couple the URL to the UI experiences in situations like this since the UI uses my full name everywhere else.  Thanks for the response so I can understand the thinking!

PR #1791, opened 2017-04-27
fix(header): Added If-None-Match header to ignore all the cache

sanbornsen:
This needs to be merged first - https://github.com/almighty/almighty-core/pull/1260

### DIFF
--- a/src/app/shared/http-service.ts
+++ b/src/app/shared/http-service.ts
@@ -40,6 +40,8 @@ export class HttpService extends Http {
     if (auth && auth.getToken() != null) {
       this.headers.set('Authorization', `Bearer ${auth.getToken()}`);
     }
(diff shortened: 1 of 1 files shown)


fabric8cd:
@sanbornsen snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1791-fabric8-planner.int.rdu2c.fabric8.io

joshuawilson:
Revert this as it is only a temp fix, the real solution is in https://github.com/almighty/almighty-core/pull/1270

joshuawilson:
We are going to try it with this still in place. Please check with @aslakknutsen before you revert this.

alexeykazakov:
https://github.com/almighty/almighty-core/pull/1270 is prod-preview now. We can revert this one in prod-preview too now.

Issue #1707, opened 2017-04-20
Parent Iteration drop down  looks distorted 


naina-verma:
Parent Iteration dropdownlooks distorted 

Expected result : 
- The drop down should collapse if use click on some other text area or other div 

Video link: https://drive.google.com/a/redhat.com/file/d/0B1qUq_laG6COM3NoWUxOZVcxRXM/view?usp=sharing

PR #1795, opened 2017-04-27
fix(iteration): fix distorted dropdown when no iteration found

vikram-raj:
fix: #1707


fabric8cd:
@vikram-raj snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1795-fabric8-planner.int.rdu2c.fabric8.io

PR #1808, opened 2017-04-27
revert setting http headers to if-none-match

joshuawilson:


fabric8cd:
@joshuawilson snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1808-fabric8-planner.int.rdu2c.fabric8.io

PR #1830, opened 2017-04-28
fix(spaces): fix crash on deselecting space.

michaelkleinhenz:
Fixes a crash when de-selecting the space. This was for example happening when switching back from the Planner to the spaces list in Platform.

### DIFF
--- a/src/app/collaborator/collaborator.service.ts
+++ b/src/app/collaborator/collaborator.service.ts
@@ -18,8 +18,12 @@ export class CollaboratorService {
   getCollaborators(): Observable<User[]> {
     return this.spaces.current.switchMap((space) => {
       // FIXME: https://github.com/fabric8-ui/ngx-fabric8-wit/issues/82
-      return this.http.get(space.links.self + '/collaborators')
-        .map(resp => resp.json().data as User[]);
+      if (space) {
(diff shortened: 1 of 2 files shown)


src/app/collaborator/collaborator.service.ts
```diff
@@ -18,8 +18,12 @@ export class CollaboratorService {
   getCollaborators(): Observable<User[]> {
     return this.spaces.current.switchMap((space) => {
```
sanbornsen: This is probably a better solution, check if this works
return this.spaces.current.filter(space => !!space)..switchMap(...)

sanbornsen: No this will not work. Never mind. It's looking good.

michaelkleinhenz: mmh..I like my solution better, easier readable.

michaelkleinhenz: ..and consistent on how we do it everywhere else.

sanbornsen: Yeah agree, go ahead. Anyway my solution is wrong.

joshuawilson:
[test]

PR #1890, opened 2017-05-17
Refactoring flatten

michaelkleinhenz:
This refactors the Planner code on a physical and logical level.


planner.module.ts
```diff
@@ -1,7 +1,7 @@
 import { NgModule } from '@angular/core';
 
-import { PlannerBoardModule } from './src/app/work-item/work-item-board/planner-board.module';
-import { PlannerListModule } from './src/app/work-item/work-item-list/planner-list.module';
+import { PlannerBoardModule } from './src/app/components/work-item-board/planner-board.module';
+import { PlannerListModule } from './src/app/components/work-item-list/planner-list.module';
```
debloper: I think we should also change `'./src/app/components/work-item-board/planner-board.module';` to `'./src/app/components/planner-board/planner-board.module';`

Reasoning behind this is "work-item" is singular entity is the planner app, and the components/modules are for the app itself. Also, consistency.

sanbornsen: @debloper Thanks. Updated.

michaelkleinhenz:
[test]

PR #1894, opened 2017-05-17
chore(planner): remove traces of "almighty" string and replace them with "planner"

### LINKED ISSUE #1891: Remove/replace references of "almighty" with planner
The term `almighty`, `Almighty`, `ALMighty` etc are referred in way too many places, and after discarding the code name long back, it doesn't make sense to carry them over.

About time we we fix it.

debloper:
closes #1891 

## also, remove the `stats.json` file
That was lingering for quite a while, without any apparent purpose or use.


fabric8cd:
@debloper snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1894-fabric8-planner.int.rdu2c.fabric8.io

PR #1921, opened 2017-05-24
fix(board): add work item iteration module

vikram-raj:
This Module got deleted while flattening refactoring. PR - #1890


sanbornsen:
We need it this way to avoid error like this 
screen shot 2017-05-29 at 4 26 56 pm

PR #1980, opened 2017-06-08
fix(detail panel): Fix duplication of work item

nimishamukherjee:
When we add a work item using the detail panel the work item gets saved but the panel does not get the newly created work item. This happens as un-subscribe gets called and the GET on the work item gets cancelled so the panel remains as is.


fabric8cd:
@nimishamukherjee snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1980-fabric8-planner.int.rdu2c.fabric8.io

nimishamukherjee:
[test]

PR #1988, opened 2017-06-09
Removed unnecessary files to deploy planner as a docker container

hectorj2f:
I believe you don't need anymore these files:
- openshift files: you won't be deployed as a single pod.
- dockerignore: there isn't any dockerfile in the repo.


kwk:
@hectorj2f @nimishamukherjee please restore the `.dockerignore` file as there are Dockerfiles still present in the repo.

PR #1994, opened 2017-06-12
Fixed dockerfiles and updated runtime/README

hectorj2f:
I kept the docker image deploy which could also be used to run planner as a docker container.


scripts/local_docker.sh
```diff
@@ -1,7 +1,6 @@
-docker build -t almighty-ui-builder -f Dockerfile.builder .
-mkdir -p dist && docker run --detach=true --name=almighty-ui-builder -e "API_URL=http://demo.api.almighty.io/api/" -t -v $(pwd)/dist:/dist:Z almighty-ui-builder
+docker build -t almighty-ui-builder -f ./deploy/Dockerfile.builder .
+mkdir -p runtime/dist && docker run --detach=true --name=almighty-ui-builder -e "API_URL=http://demo.api.almighty.io/api/" -t -v ./runtime/dist:/dist:Z almighty-ui-builder
```
kwk: Since when do volume mounts work with `./` instead of a full path?

kwk:
@hectorj2f  What was the reasoning for adding the GPG keys to the repo? To increase speed or avoid problems with key-servers?

scripts/local_docker.sh
hectorj2f: fixed, it got into the wire by mistake. my bad

hectorj2f:
@kwk I kept them as they were before. This operation is also performed by the nodejs container at this moment.

PR #2024, opened 2017-06-21
Sass to less

SMahil:
***WIP***


mindreeper2420:
@SMahil was this started after the flex work was merged? I noticed a lot of previously removed flex styles in the files that were converted to Less.

SMahil:
@mindreeper2420 No it was started after flex changes but somehow changes are not there.

PR #2084, opened 2017-07-24
fix(functests) stick to chrome driver 2.30 version

hectorj2f:
As you can see in the job activity of planner PRs, all are failing in the funciontal tests cause a new version of the chrome driver is used 2.31. This version is most certainly the cause of this problem, so we'll stick to 2.30 version as done in the past.


fabric8cd:
@hectorj2f fabric8/fabric8-ui:SNAPSHOT-PR-2084-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2084-fabric8-planner.badger.fabric8.io

PR #2121, opened 2017-08-29
Containerized Planner, composed with DB and API

debloper:
The setup is functional and resolves quite a few corner cases but has some limitations.
Authentication integration works with KeyCloak's `fabric8-test` realm.

To run/verify this PR, try the following:
```sh
docker build -t fabric8-planner-runtime .
docker-compose up
```
You should be able to visit `http://localhost:8088/` on your browser at this point. This expects your system to have docker (and docker engine v1.13.0+ i.e. with version v3.x compose file support) set up and the service running. When done, do the following to clean up:
```sh
docker-compose down
docker images # lists the images
docker rmi <imageID>
```
Alternatively, you can also run _only_ planner image, against external (prod/demo) API sources. In that case, use something like this:
```sh
docker build -t fabric8-planner-runtime .
docker run -it --name=fabric8-planner-runtime -p 8088:8080 \
        -e "API_URL=http://demo.api.openshift.io/api/" \
        -v $(pwd)/runtime/dist:/dist:Z fabric8-planner-runtime
```
If you just want to run the unit-tests (or some other script in the container, instead of starting the server), then please uncomment and edit the `services::planner::command` field in the `docker-compose.yml`.

### Known Limitations
- Uses integrated runtime, so please don't expect full fledged OSIO workflow
- Data won't persist between subsequent runs - it's more of a choice than a limitation


Dockerfile
```diff
@@ -62,11 +62,16 @@ ENV WORKSPACE=$HOME/fabric8-planner
 RUN mkdir $WORKSPACE
 
 COPY . $WORKSPACE
-RUN chown -R ${FABRIC8_USER_NAME}:${FABRIC8_USER_NAME} $HOME/*
```
sanbornsen: Why the user is removed? Will the app be run inside container using root user?

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2121-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2121-fabric8-planner.badger.fabric8.io

docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      - core
+    networks:
+      - default
+    ports:
+      - "8088:8080"
+    volumes:
+      - ./runtime/dist:/dist:Z
+  core:
+    image: fabric8/fabric8-wit:latest
+    command: -config /usr/local/wit/etc/config.yaml
```
sanbornsen: Where is this file?

Dockerfile
debloper: The user isn't removed, this is just forcing recursive file permissions (ownership) onto the created user.
This avoids too many corner cases (and cuts down build time by 5min).

Unless we have very good reason, I'm sure creating and switching to another user only adds complexity in the setup. Using `root` user works perfectly fine in this sandboxed environment.

docker-compose.yml
debloper: `https://github.com/fabric8-services/fabric8-wit/blob/master/config.yaml`

nimishamukherjee:
When I run docker-compose up I get:
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/

debloper:
@nimishamukherjee you might be using older version of docker-compose (that doesn't support compose file version 3.x). We're using compose file version 3 here because it adds supports for docker's swarm mode (might be useful for us going forward), and reduces complexities in multiple other directives.

I'll add a note in the doc mentioning the docker-compose version requirement.

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2121-2 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2121-fabric8-planner.badger.fabric8.io

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2121-2 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2121-fabric8-planner.badger.fabric8.io

debloper:
Updates to this PR regarding the AUTH implementation I've been smashing my head around:

- KeyCloak authentication is harder to implement without using full platform, it also has several gotchas regarding the hostname, redirection URL etc. and often misbehaves. For automated testing, it doesn't bring enough value to have it implemented, for the work it needs to get there. Having a conversation with @michaelkleinhenz we're going to use single user mode for the Docker Compose implementation, in order to allow write operations.
- Hence this pushes back the AUTH implementation to minishift based orchestration, but for now allows the new E2E test suits to run in an orchestrated and sandboxed (docker-compose) environment.
- Running WIT/core in auth-less (token based) dev environment isn't readily feasible for running as a containerized API, and it need some works as explained in fabric8-services/fabric8-wit#1630

I suggest we go ahead with merging this patch as it works quite well to run the setup as we want it, and the missing auth-piece needs works mostly on the other repo in order to be available. So, instead of blocking ourselves on that, let's add it on top of this when we're unblocked.

debloper:
KeyCloak SSO is now integrated, but the client needs to identify itself for user-auth to work.
App will be on `/login` page loop because of:
```
oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"unauthorized_client","error_description":"Invalid client secret"}
```

docker-compose.yml
sanbornsen: So is this a path inside the container?

debloper:
Update on the last commit:

- Turns out the config file was messing up the authentication by resetting the KC realm
- Running dev-mode with basic defaults works as intended and makes the auth go through
- It does mean that we're not having the performance monitoring scripts running in this setup
- But in practice, that's okay and is even better, by cutting down unnecessary complexity & code

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2121-2 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2121-fabric8-planner.badger.fabric8.io

michaelkleinhenz:
@sanbornsen can you re-check your review here please..

michaelkleinhenz:
@debloper please rebase.

rgarg1:
@debloper  I ran into the same issues that Nimisha was facing. Are you using docker-ce? I tried installing that, but then I can't run docker command from the terminal. Am I missing something?

debloper:
@rgarg1 I wasn't using Docker CE, just the regular package installed with DNF. Please check the main description about the pre-req:

> This expects your system to have docker (and docker engine v1.13.0+ i.e. with version v3.x compose file support) set up and the service running.

You need to ensure this requirement is met.

rgarg1:
@debloper I met the reqs with Docker CE and got the planner running.

But, I got stuck at this page. I tried my OSIO credentials as well as prod-preview, but none work. How do I get past this stage?
login page

debloper:
@rgarg1 please get your credentials approved for fabric8-test realm, that's the only breakpoint that seems likely in this scenario. @sbose78 should be able to help you get the account approved.

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2121-6 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2121-fabric8-planner.badger.fabric8.io

ldimaggi:
Just tried this:

docker build -t fabric8-planner-runtime .
docker-compose up

But - only got this error in trying to log in:

screenshot 66

PR #2138, opened 2017-09-08
Re-Structure less files , classnames

SMahil:
This PR includes:

Restructure LESS files for better structure
Rename classnames according to component name
Add Prefix f8- to classnames
Update styles to match guidelines


SMahil:
#2139

PR #2151, opened 2017-09-14
Replace hardcoded localhost

ldimaggi:
Replaced references to hardcoded localhost with test parameter (browser.params.target.url) for target server URL - for example:
npm run test:funcsmoke http://localhost:8088


runtime/src/tests/app.component.spec.js
```diff
@@ -1,10 +1,12 @@
 describe('Splash page', function() {
 
   beforeEach(function() {
-      browser.get('http://localhost:8088/');
+      browser.get(browser.params.target.url + "/');
```
rgarg1: String quote mismatch" " closed with '

runtime/tests/run_functional_tests.sh
```diff
@@ -46,9 +46,9 @@ echo done.
 echo Running tests...
 if [ -z "$1" ]
   then
-    ../node_modules/protractor/bin/protractor tests/protractor.config.js 2>&1
+    ../node_modules/protractor/bin/protractor tests/protractor.config.js --params.target.url=$1 2>&1
```
rgarg1: One more ref to localhost at line 42:
https://github.com/ldimaggi/fabric8-planner/blob/91432f80be713c9259b6b41f3d731b107d921a07/runtime/tests/run_functional_tests.sh#L42

rgarg1:
@ldimaggi All tests passed locally. Now the big next step is to make the test scripts work with the new DB. Let's talk about that.

PR #2158, opened 2017-09-19
fix(labels): User experience bugs fixed

sanbornsen:
- Save labels on close selector dropdown
- Show loading while resolvong or saving labels

PR #2165, opened 2017-09-25
fix(dynamic-field): remove labels from dynamic field

vikram-raj:


vikram-raj:
Fixed in https://github.com/fabric8-ui/fabric8-planner/pull/2158

PR #2281, opened 2017-10-16
fix(WI-quick-add): Remove redundant scrollbar from description

jarifibrahim:
This PR fixes https://openshift.io/openshiftio/openshiftio/plan/detail/1550

### DIFF
--- a/src/app/components/work-item-quick-add/work-item-quick-add.component.less
+++ b/src/app/components/work-item-quick-add/work-item-quick-add.component.less
@@ -40,10 +40,12 @@
     &-story {
       padding-right: 70px;
       padding-left: 42px;
-      textarea[class='f8-quickadd-desc'] {
-        max-height: 100px;
-        height: 26px !important;
-        padding-top: 0;
+      textarea.f8-quickadd-desc {
+        bottom: -25px;
+        max-height: 200px;
+        max-width: 407px;
+        position: absolute;
+        resize: none;
       }
     }
     &-btn {
@@ -94,10 +96,11 @@
     margin-left: 0;
     background: @color-pf-white;
     .Border(2px, solid, @color-pf-black-400);
-    textarea[class="f8-quickadd-desc"] {
+    textarea.f8-quickadd-desc {
       width: 100%;
       height: 29px;
       margin-top: 5px;
+      resize: none;
     }
     &-btn-add {
       margin-top: 5px;
@@ -124,6 +127,6 @@
 }
 @-moz-document url-prefix() {
   .f8-quickadd__wiblk-story {
-    textarea[class="f8-quickadd-desc"] { max-height: 26px !important; }
+    textarea.f8-quickadd-desc { max-height: 26px !important; }
   }
 }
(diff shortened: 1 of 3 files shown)


src/app/components/work-item-quick-add/work-item-quick-add.component.less
```diff
@@ -40,10 +40,12 @@
...
       padding-right: 70px;
       padding-left: 42px;
-      textarea[class='f8-quickadd-desc'] {
-        max-height: 100px;
-        height: 26px !important;
-        padding-top: 0;
+      textarea.f8-quickadd-desc {
+        bottom: -25px;
+        max-height: 200px;
+        max-width: 407px;
```
SMahil: max-width is not needed

src/app/components/work-item-quick-add/work-item-quick-add.component.less
```diff
@@ -40,10 +40,12 @@
...
+        position: absolute;
```
SMahil: why position absolute?

src/app/components/work-item-quick-add/work-item-quick-add.component.less
```diff
@@ -94,10 +96,11 @@
     margin-left: 0;
     background: @color-pf-white;
     .Border(2px, solid, @color-pf-black-400);
-    textarea[class="f8-quickadd-desc"] {
+    textarea.f8-quickadd-desc {
```
SMahil: why change of syntax?

src/app/components/work-item-quick-add/work-item-quick-add.component.less
```diff
@@ -124,6 +127,6 @@
 }
 @-moz-document url-prefix() {
   .f8-quickadd__wiblk-story {
-    textarea[class="f8-quickadd-desc"] { max-height: 26px !important; }
+    textarea.f8-quickadd-desc { max-height: 26px !important; }
```
SMahil: why change of syntax

src/app/components/work-item-quick-add/work-item-quick-add.component.less
jarifibrahim: screenshot from 2017 10 17 00 24 31

^^This is why max-width is necessary. The description field is a `col-md-6` and title field is also a `col-md-6` Without max-width, the description field overlaps with the buttons.

src/app/components/work-item-quick-add/work-item-quick-add.component.less
jarifibrahim: We want the description field to grow in vertically upward direction. That's why it's position has to absolute. Without this, the textarea will grow vertically downward.

src/app/components/work-item-quick-add/work-item-quick-add.component.less
jarifibrahim: `textarea[class='f8-quickadd-desc']` isn't the correct way of selecting a textarea with class f8-quickadd-desc. If you use `textarea[class='f8-quickadd-desc']` none of the styles are reflected.

SMahil:
Textarea is overflown from quick add container

screen shot 2017-10-17 at 12 55 47 pm

jarifibrahim:
@SMahil This is the expected result, right? We want the textarea to grow vertically upward. Textarea has to overflow from quick-add-container, otherwise it won't grow :laughing:

SMahil:
I don't think this is expected behaviour.It seems broken now. We can approach to UX team for better solution

jarifibrahim:
Closing this PR since it does not fix the actual issue and rather breaks the UI.

PR #2308, opened 2017-10-25
fix(test):refactor tests for vertical test setup

rgarg1:
@ldimaggi Running into angular sync issues as the login page is non-Angular. Here is the error:

- Failed: Timed out waiting for asynchronous Angular tasks to finish after 120 seconds. This may be because the current page is not an Angular application. Please see the FAQ for more details: https://github.com/angular/protractor/blob/master/docs/timeouts.md#waiting-for-angular

Login only works with ignoreSynchronization flag if set to true, but the tests work with the flag set to false. I tried disabling and the re-enabling the ignoreSynchronization, but run into the above errors. 

Your help needed.

### DIFF
--- a/runtime/src/tests/app.component.spec.js
+++ b/runtime/src/tests/app.component.spec.js
@@ -1,10 +1,10 @@
 describe('Splash page', function() {
 
   beforeEach(function() {
-      browser.get('http://localhost:8088/');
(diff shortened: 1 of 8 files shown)


rgarg1:
Not needed anymore as we've new tests that bypass Angular stability issues.

PR #2321, opened 2017-11-02
fix(test): tokenized tests for vertical testing

rgarg1:
The tests now launch successfully and the annoying login screen is also bypassed, however the tests still do not execute due to ngZone clicks making the protractor wait indefinitely.


rgarg1:
Not needed anymore as we've new tests that bypass Angular stability issues.

PR #2350, opened 2017-11-28
fix(workItem): Fix smoketest warnings

jarifibrahim:
This PR fixes the warnings shown below
download 4

### DIFF
--- a/runtime/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
+++ b/runtime/src/tests/work-item/work-item-list/page-objects/work-item-list.page.js
@@ -282,7 +282,7 @@ class WorkItemListPage {
   }
 
   clickWorkItem(workItemElement) {
-    workItemElement.element(by.css(".f8-wi__list-description")).element(by.css("p")).click()
+    workItemElement.$$(".f8-wi__list-description").first().element(by.css("p")).click()
     return new WorkItemDetailPage();
   }
 


src/app/components/work-item-list-entry/work-item-list-entry.component.html
```diff
@@ -46,14 +46,14 @@
...
-            <div class="f8-wi__list-description">
+            <div class="f8-wi__list-labels">
               <f8-label [labels]="workItem.relationships?.labels?.data ?
                 workItem.relationships?.labels?.data : []"
                 [truncateAfter]='4'
                 [allowDelete]="false"
                 (onLabelClick)="labelClick($event)"></f8-label>
             </div>
-            <div class="f8-wi__list-description width-100px">
+            <div class="f8-wi__list-status width-100px">
```
SMahil: Why you changed the class names?

jarifibrahim: Becuase they are incorrect. This div contains wi status. Also, protractor tests were throwing warnings as there were 3 `f8-wi__list-description` for each workItem

SMahil: Yes it contains several things such as title, label and status but there are certain common styles set to those divs through that one class. Moreover I choose the name description because it describes the detail of that particular list item. 

You can't change the classnames like this as the styles associated to it are not going to apply now

jarifibrahim: This means we'll have to add IDs to separate description/status/labels. The tests need a unique selector to click on the work item title.

SMahil: Then add Ids but don't change classes

jarifibrahim: Either we add unique classes to description/label/status or we'll have to change the CSS.

jarifibrahim: @SMahil We have multiple workitems and each work item has a description. We cannot use IDs in this case

SMahil: No Dont change CSS. AFA adding ids to label and status, I dont know if you are checking click item on that but they are not clickable

SMahil: in that case add extra class for the item which is going to be clicked

jarifibrahim: I'll add a new class for the clickable div :+1:

sthaha: > The tests need a unique selector to click on the work item title.

@jarifibrahim not really ... you can make use `$$` or `element.all` and filter on those to obtain the unique item. There is also [cssContainingText](http://www.protractortest.org/#/api?view=ProtractorBy.prototype.cssContainingText) api that you can make use of. Have you considered these alternatives?

jarifibrahim: Thanks @sthaha @SMahil. I've updated the PR.

fabric8cd:
@jarifibrahim fabric8/fabric8-ui:SNAPSHOT-PR-2350-2 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2350-fabric8-planner.badger.fabric8.io

fabric8cd:
@jarifibrahim fabric8/fabric8-ui:SNAPSHOT-PR-2350-3 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2350-fabric8-planner.badger.fabric8.io

PR #2359, opened 2017-12-11
fix(broadcaster): remove broadcaster from provider list

dgutride:
This is breaking notifications on the platform because it's created a second service (and events can't be delivered to any  listening component once these modules have been loaded).


fabric8cd:
@dgutride fabric8/fabric8-ui:SNAPSHOT-PR-2359-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2359-fabric8-planner.badger.fabric8.io

PR #2150, opened 2017-09-13
fix(stylelint): add stylelint to build process

mindreeper2420:
Add Stylelint to the build process, with warnings and errors displayed as necessary

Build process configured so that it does not fail on error, but will display all errors and/or warnings in the console.

closes https://github.com/fabric8-ui/fabric8-planner/issues/2147

Issue #2147, opened 2017-09-13
Add Stylelint to Repo


mindreeper2420:
In order to match the CSS code guidelines, Stylelint must be enabled to check for CSS code quality. Stylelint is currently enabled in fabric8-ui and ngx-widgets, with the update to fabric8-recommender in a PR.

The configuration file to use can be found here: https://github.com/fabric8-ui/fabric8-ui/blob/master/.stylelintrc

PR #2372, opened 2017-12-18
fix(stylelint): add stylelint to build process

mindreeper2420:
add Stylelint to the build process, with warnings and errors displayed as necessary

BREAKING CHANGE: build, if stylesheets are not written to spec

https://github.com/fabric8-ui/fabric8-planner/issues/2147

replaces original PR: https://github.com/fabric8-ui/fabric8-planner/pull/2150


.stylelintignore
```diff
@@ -0,0 +1,4 @@
+# Ignore guidelines for stylelint
+# These are to be written in the same format as gitignore
+# Any directories or files are automatically found by the stylelint process
+
```
SMahil: we need to ignore formulas.less

joshuawilson:
ping me if you need help on this one

SMahil:
@mindreeper2420 This PR fails locally also. Its failing on "copy-watch-all" task

gulpfile.js
```diff
@@ -14,14 +15,19 @@ var gulp = require('gulp'),
...
   cp = require('child_process'),
-  exec = require('gulp-exec'),
+  del = require('del'),
+  exec = require('gulp-exec').exec,
+  KarmaServer = require('karma').Server,
+  lessCompiler = require('gulp-less'),
+  stylelint = require ('gulp-stylelint'),
+  cssmin = require ('gulp-cssmin'),
   ngc = require('gulp-ngc'),
-  changed = require('gulp-changed'),
```
jarifibrahim: @mindreeper2420 The gulp build is failing with `changed` not found.  `gulp-changed` is used at https://github.com/mindreeper2420/fabric8-planner/blob/1f4abc328cc4ec741d065d609cc0844ca43f74ab/gulpfile.js#L52

mindreeper2420:
I'll be updating this in the next day or so. Please do not merge master into this PR - I will do so locally and push an update as necessary.

gulpfile.js
mindreeper2420: fixed - reference re-added.

mindreeper2420:
@SMahil the build fails as there are stylelint errors.

If any command includes `gulp stylelint`, then the build will fail, as **failAfterError** is set to **true**.

If you run a command that utilizes `gulp lint-less`, then that command will not fail, as **failAfterError** is set to **false**.

You can turn off the fail on error, but then the build will still run and pass even with style issues. As noted in the command, this PR breaks Planner until these are fixed.

joshuawilson:
@mindreeper2420 can you change the PR to submit the linting but not fail and then a new PR with the failing turned on. This will allow them to turn it on if they want and make it an issue to address or not.

mindreeper2420:
@joshuawilson updated to remove failure on build errors, changed all errors alerts to be warnings and added a combiner to allow configuration of showing all errors in one grouping, rather than by line/file. I've added comments in the gulpfile where **false** needs to be set to **true** to turn build failures back on.

mindreeper2420:
@joshuawilson @SMahil I pushed another set of changes, fixing files that had to be fixed in order to comply with stylelint. The task `gulp build:library`, which is required for `gulp copy-watch-all` now runs and completes.

I had to add a unit of measurement to the unit-whitelist setting, as I don't know what its intention actually is (the unit is `*pf-custom-navbar-item-height`). This should be fixed and removed from the whitelist in the future.

joshuawilson:
@michaelkleinhenz it is very sad that @mindreeper2420 has tried twice now to get linting into the planner and it has been ignored.

PR #2379, opened 2017-12-25
fix(script): Bootstrap.sh can now generate fabric8-ui docker image

jarifibrahim:
A user can generate `fabric8-ui` docker image with latest planner source using `bootstrap.sh` script

How to run - 
`./bootstrap.sh PATH_TO_FABRIC8_UI`

Result - 
`fabric8-planner-platform` docker image is built.
To run the image `docker run -e PROXY_PASS_URL="https://api.free-int.openshift.com" -p 8080:8080 fabric8-planner-platform`

NOTE - The script assumes that all the required node modules are already installed and won't install any packages.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -1,119 +1,106 @@
+##
+# Bash script to generate fabric8-ui image with latest planner source
+# NOTE - The script assumes the user has all the required node modules in fabric8-planner and
+#        fabric8-ui directory.
+##
 #!/bin/bash
-set -x
+set -e # exit immediately when a command fails
 
-echo "Setting up working directories..."
-BASEPATH=`pwd -P`
-PLANNERPATH=""
-PLATFORMPATH=""
-DRYRUN=false
+build_and_install_planner() {
+    print_success "STEP 1/4 - Building planner library..."
+    cd $PLANNER_PATH
+    npm run build:library || {
+        print_err "Failed to build fabric8-planner. Please fix the error and try again."
+        exit 1
+    }
+    # Pack fabric8-planner into fabric8-planner@0.0.0-development.tar.gz
+    package_name=$(npm pack dist)
+    print_success "STEP 1/4 - Building planner library... - Done"
+    install_planner "$package_name"
+}
 
-echo "Parsing CLI arguments..."
-for i in "$@"
-do
-case $i in
-    --planner=* )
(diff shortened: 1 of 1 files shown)


fabric8cd:
@jarifibrahim fabric8/fabric8-ui:SNAPSHOT-PR-2379-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2379-fabric8-planner.badger.fabric8.io

pranavgore09:
I have not tested this yet but with this support and a small change in here https://github.com/fabric8-services/fabric8-wit/pull/1817/files#diff-3981d9a6f3c5229eb93272a67600b446R12 will be good to take care of using that image in minishift :smile:

jarifibrahim:
@pranavgore09 Thanks. Instead of `fabric8-planner-platform` we could call this as `fabric8-ui` with a custom tag. Maybe something like `fabric8-ui:test` image. This way we can directly use this in the minishift setup.

fabric8cd:
@jarifibrahim fabric8/fabric8-ui:SNAPSHOT-PR-2379-2 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2379-fabric8-planner.badger.fabric8.io

debloper:
As talked about in mattermost, the purpose of this PR isn't very clear.

> Bootstrap.sh can now generate fabric8-ui docker image
bootstrap.sh already does it, in fact it does so in a contained environment and not on the host like in this PR. It also uses configurable defaults instead of rigid assumptions of parameters.

So, please clarify first what's happening in here, before we get into the review.

jarifibrahim:
@debloper The motive behind this PR is to simplify the existing `bootstrap.sh` script. This PR generates the `fabric8-ui` image similar to how CI/CD builds the image.
The assumption that `fabric8-ui` and `fabric8-planner` have all the dependencies installed is based on the fact that if a developer is "developing", he/she would have all the dependencies. 

Also, I removed the `dryrun` flag since we do not need it anymore. The `dryrun` flag was added to test the orchestration but it doesn't actually test the orchestration. It tests if nginx can render the index.html page. We can't conclude that the orchestration is working by the image generated through `dryrun`. All we can conclude is that there is a basic nginx image which can render the index.html file.

jarifibrahim:
Closing this PR since we already have a `bootstrap.sh` script and this script doesn't actually change anything apart from the script structure.

PR #2413, opened 2018-01-19
fix(label-service): Add check for null current space. Fixes SEV1 issue1928

nimishamukherjee:
fix(label-service): Add check for null current space. Fixes https://github.com/openshiftio/openshift.io/issues/1928

### DIFF
--- a/src/app/services/label.service.ts
+++ b/src/app/services/label.service.ts
@@ -32,11 +32,15 @@ export class LabelService {
    */
 
   getLabels(): Observable<LabelModel[]> {
-    return this.spaces.current.switchMap(
-      currentSpace => this.http.get(currentSpace.links.self + '/labels')
-    ).map (response => {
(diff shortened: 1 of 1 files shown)


bartoszmajsak:
Some regression test for this bug would be nice.

fabric8cd:
@nimishamukherjee fabric8/fabric8-ui:SNAPSHOT-PR-2413-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2413-fabric8-planner.badger.fabric8.io

rgarg1:
LGTM. The issue is no longer reproducible.

However, an observation that I want to record here:

Sometimes, clicking on the 'opt in' button or the 'profile' link, takes the user back to the dashboard, instead of the user profile page. This happens only sometimes.

The fix has no ill-effect on the Planner labels functionality and its working as expected.

PR #2431, opened 2018-02-06
fix(CI/CD): Run TS tests instead of JS tests in CD step

jarifibrahim:


jarifibrahim:
@pranavgore09 I don't think we can test this without merging. We could try to replace the `pipeline.ci()` with `pipeline.cd()` in the Jenkins file but then it would push all the new packages to npmjs.org and dockerhub.

fabric8cd:
@jarifibrahim fabric8/fabric8-ui:SNAPSHOT-PR-2431-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2431-fabric8-planner.badger.fabric8.io

PR #2474, opened 2018-02-21
chore(npm): remove unnecessary NPM modules from dev-deps

debloper:
Trim down dev-deps by 50% (from 93 to 47)

BEFORE
```
added 1895 packages in 27.407s
```

AFTER
```
added 1317 packages in 19.969s
```


fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2474-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2474-fabric8-planner.badger.fabric8.io

pranavgore09:
@debloper I really can not say yes or no to such changes! but build passed so it should be fine..
@jarifibrahim  ?

michaelkleinhenz:
I am happy about reducing dependencies, but we need to be absolutely sure that this does work. I'd like to see the tests completed (or at least in a reasonable state) and see if this is working before we can do that. We have a lot of potential sideeffects here.

debloper:
> @debloper as we are nearing GA, we cannot play around with features which break and block us completely.

And this is why I suggested the **separation of Upstream and Downstream**.

There's nothing inherently wrong with either prioritizing improvement vs prioritizing delivery, but when it's either/or it clashes and not in a very good way. Separate them so we can get better at both, instead of being worse at both, trying to make compromises which is the current scenario.

It's not even a new concept to introduce (at least in Red Hat). Fedora/RHEL, Gluster/RHS etc. are good examples of that. Why can't we adopt similar approach that doesn't tie our hand from trying anything new/better/different in the name of "can't risk delivery"? Those two should be separate concerns.

jarifibrahim:
I agree we need this change but changing it right now might not be a good idea. We could end up breaking *everything*.

sanbornsen:
I love these changes. Let all of us pull it locally and test every single corner. If it's all good, then this should go in.

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2474-2 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2474-fabric8-planner.badger.fabric8.io

debloper:
In order to state what seemed obvious (to me, and a few others here)... this patch deals with dev-deps, and has no feasible reason to "break and block" application _features_. It will either build, or it wouldn't - depending on whether it has just enough modules to build the project, or not.

The only reason for someone to assume on that line, points towards unfamiliarity and lack of understanding of how `Node.js` and `npm` ecosystem works. Earlier it seemed a bit harsh to put it this way, but that seems to do no good.

Feel free to merge/reject the PR, for I'll discontinue maintaining/updating it.

joshuawilson:
Closing this as it will be easier to clean it up in a new PR then to update this one in a rebase.

PR #2476, opened 2018-02-22
fix(labels): update the WI-label association callout to not confuse with label creation

debloper:
Closes openshiftio/openshift.io#1836

### DIFF
--- a/src/app/components_ngrx/label-selector/label-selector.component.html
+++ b/src/app/components_ngrx/label-selector/label-selector.component.html
@@ -3,7 +3,7 @@
   [toggleButtonRef]="toggleButtonRef"
   [dropdownItem]="dropdownItem"
   [dropdownFooter]="dropdownFooter"
-  [headerText]="'Add label'"
+  [headerText]="'Assign Label(s)'"
   [menuItems]="labels"
   [showSearch]="backup.length > 8"
   (onSelect)="onSelect($event)"
@@ -15,7 +15,7 @@
   <ng-template #toggleButtonRef>
     <span class="clickable add-label" (click)="closeAddLabel()">
       <i class="pficon pficon-add-circle-o"></i>
-      Add label
+      Assign Label(s)
     </span>
   </ng-template>
   <ng-template #dropdownItem let-color="color" let-border="border" let-name="name">


fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2476-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2476-fabric8-planner.badger.fabric8.io

src/app/components/label-selector/label-selector.component.html
```diff
@@ -13,7 +13,7 @@
   <ng-template #toggleButtonRef>
     <span class="clickable add-label" (click)="closeAddLabel()">
       <i class="pficon pficon-add-circle-o"></i>
-      Add label
+      Assign labels
```
nimishamukherjee: @Preeticp is it Assign label or Assign Label?

Preeticp: Ideally, I would suggest Add label, but we have Title case being used in majority of the cases in OSIO, so to keep it consistent let's use Assign Label.

Preeticp: If it has to be changed it should be across OSIO as a whole to sentence case.

nimishamukherjee: Thanks @Preeticp 
@debloper please make it "Assign Label"

debloper: Please reflect whether it should be `Label` or `Labels` cause it deals with multiple association.

rgarg1: @Preeticp So should it be "Assign Labels" ? Can you confirm.

rgarg1:
Its still "Assign labels" in the snap-shot.

debloper:
> Its still "Assign labels" in the snap-shot.

@rgarg1 please check my comment in the review thread. I haven't gotten a response in favor/against the pluralization from @Preeticp

debloper:
It's better if the confirmation comes from @Preeticp, and as she has amended it to `Assign Label(s)` keeping multiple association in mind (and not only casing, which was the previous suggestion about).

The updated suggestion seems good to me.

sanbornsen:
This change has to be done in the new version. You can find the relative file in **components_ngrx** directory.

debloper:
@sanbornsen I don't see the change reflected. It's still showing `Add label` to me on http://fabric8-ui-fabric8-planner.badger.fabric8.io

screenshot from 2018-03-12 15-06-31

Also, fixing issues that others are already working on without notifying them or on the issues that someone else is taking it over (this is the 4th PR I see your comment, mentioning same thing) is not the best use of time for both the parties. Please notify before hands if you absolutely have to take over an already WIP/submitted/in-review PR.

nimishamukherjee:
@debloper add this change to the ngrx-component.

rgarg1:
@debloper The primary option in the workitem preview is fixed and showing as __Assign Label(s)__ however, there is another _Add Label_ label inside the drop-down that needs to be addressed. See snapshot attached.

add label

rgarg1:
LGTM @debloper

PR #2493, opened 2018-03-08
fix(label): error on duplicate label creation is handled

sanbornsen:
This PR fixes the following issue 

https://github.com/openshiftio/openshift.io/issues/2469


fabric8cd:
@sanbornsen fabric8/fabric8-ui:SNAPSHOT-PR-2493-1 is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2493-fabric8-planner.badger.fabric8.io

sanbornsen:
@Raunak1203 all the error message are not specific at this moment. We have no tech in place to handle error codes and show a relevant message. That should be another story. 
The concept of enabling the button is when you start editing the system assumes you are putting another name. What do you think a better experience could be? I'll update according to your suggestion.

Raunak1203:
@sanbornsen for enabling the create label Button , My suggestion is if label starts with space or special characters then it should not be enabled. WDYT?

alien-ike:
### Ike Plugins (test-keeper)

Thank you @sanbornsen for this contribution!



If you are an admin and you are sure that no test is needed then you can use the command `/ok-without-tests` as a comment to make the status green.


For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin - for example exclude certain file types so if PR contains only them it won't be checked.

nimishamukherjee:
Created https://github.com/openshiftio/openshift.io/issues/3117 to address
> for enabling the create label Button , My suggestion is if label starts with space or special characters then it should not be enabled. WDYT?

nimishamukherjee:
[test]

PR #2494, opened 2018-03-08
fix(iteration): allow pressing enter on iteration name to create/update iteration

debloper:


sanbornsen:
This PR is not relevant as per new planner updates. Hence closing it.

PR #2526, opened 2018-04-04
Fix docker cmd

pranavgore09:
Remove unwanted ENV, add `-it` so that we can kill the build with ctrl+c
new format looks like

```diff
+ ========= Run snapshot by running following command =========
+ docker run -it -p 5000:8080 registry.devshift.net/fabric8-ui/fabric8-planner:SNAPSHOT-PR-2526-22c271
+ =============================================================
```

notice the color and image tag.


cico_run_tests.sh
```diff
@@ -108,5 +108,5 @@ docker push ${REGISTRY}/${IMAGE_REPO}:${TAG}
 PULL_REGISTRY="registry.devshift.net"
 image_name="${PULL_REGISTRY}/${IMAGE_REPO}:${TAG}"
 echo "======= Snapshot can be created by running following command"
-echo "docker run -e PROXY_PASS_URL=\"https://api.free-stg.openshift.com\" -p 8080:8080 ${image_name}"
+echo "docker run -it -p 5000:8080 ${image_name}"
```
jarifibrahim: Why do we need `-it` ?

pranavgore09: to be able to kill it using `ctrl+c`

PR #2533, opened 2018-04-09
CICO: Remove commit hash from snapshot image tag

jarifibrahim:
This PR fixed the snapshot image tag.

Since there are limitations to what we can show via CICO success comment[0], we need to remove the commit hash from the image tag.

[0] - https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml#L335

### DIFF
--- a/cico_run_tests.sh
+++ b/cico_run_tests.sh
@@ -111,9 +111,9 @@ image_name="${PULL_REGISTRY}/${IMAGE_REPO}:${TAG}"
 set +x
 # Pretty print the command for snapshot
 echo
-echo -e "\e[92m========= Run snapshot by running following command =========\e[0m"
-echo -e "\e[92m\e[1mdocker run -it -p 5000:8080 ${image_name}\e[0m"
-echo -e "\e[92m=============================================================\e[0m"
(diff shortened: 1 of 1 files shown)


cico_run_tests.sh
```diff
@@ -96,10 +96,7 @@ else
 fi
 
 # Build and push image
-# Use default length when not provided
-echo "${DEVSHIFT_TAG_LEN:=6}"
-VERSION_NUMBER=$(echo $GIT_COMMIT | cut -c1-${DEVSHIFT_TAG_LEN})
-TAG="SNAPSHOT-PR-${ghprbPullId}-${VERSION_NUMBER}"
+TAG="SNAPSHOT-PR-${ghprbPullId}"
```
pranavgore09: what is the limitation on cico for tagging?
and instead of removing the tag I would like to use 2 or 3 characters from the hash. we need tag each new build on new commit.

jarifibrahim: Yes, agreed. But we cannot show custom tags on success comments. The CICO build uses the static success message.

jarifibrahim:
[test]

PR #2529, opened 2018-04-06
fix(iteration): fix user experience for iteration block and icon

SMahil:
### [ ] What does this PR do? _[Mandatory]_
          Fix User Experience for Iteration block

### [ ] What issue/task does this PR references? _[Mandatory]_
https://github.com/openshiftio/openshift.io/issues/2652

### [ ] Are the tests Included? _[Mandatory]_
NA

#### [ ] Is the documentation Included?
NA

#### [ ] Are the Release Notes included?
NA

#### [ ] Is/Are the @mention(s) to reviewer(s) included
@nimishamukherjee @rgarg1

PR #2564, opened 2018-04-17
fix(iteration): make the visual & programmatic click area for selecting iterations to be identical

debloper:
From: 

To: screenshot from 2018-04-10 15-13-33

### [ ] What issue/task does this PR references? _[Mandatory]_
Closes openshiftio/openshift.io#2970

### [ ] Are the tests Included? _[Mandatory]_
N/A

#### [ ] Is the documentation Included?
N/A

#### [ ] Are the Release Notes included?
N/A

#### [ ] Is/Are the @mention(s) to reviewer(s) included
Yes


SMahil:
This behavior is not intended and also it is not UX approved.

kwk:
> This behavior is not intended and also it is not UX approved.

@SMahil well, what is intended then? What if I have an iteration of one character "a". Shall the clickable area be only the little letter "a"? I doubt that this is thought through.

SMahil:
@kwk Intended behavior (Quick solution)is to click iteration name for now. Single character iteration name is not real life use-case. Currently, UX is considering all these edge cases and provide the better experience covering all edge scenarios.

kwk:
> @kwk Intended behavior (Quick solution)is to click iteration name for now. Single character iteration name is not real life use-case. Currently, UX is considering all these edge cases and provide the better experience covering all edge scenarios.

@SMahil When can we expect UX to come up with a proposal?

SMahil:
@kwk yes, his proposal is being considered. Further changes are there and hopefully, we can close it tomorrow

debloper:
@SMahil I did not receive any update, so I'm not entirely sure if the proposal was approved/rejected by UX. However, I do see a very similar approach (DOM and interaction changes) has been used in updated #2529 (which was just cursor change previously).

If the suggested UX is acceptable/approved, why do we need to work on redundant PRs?

SMahil:
@debloper I fixed it in below PR https://github.com/fabric8-ui/fabric8-planner/pull/2529

debloper:
> I fixed it in below PR #2529

Not quite clear what you meant by that.

My point being, the fix was already submitted in this PR. Why do we have to copy it over to another PR? The only part that's different in that PR is using a non-semantic `<div>` instead of `<a>` to contain the link, and then emulating a link behavior by forcing cursor change.

Is there a reason for this PR to not satisfy the requirements? If not, then you're marked as a reviewer, please share your suggestion. I don't think it's quite justified to make redundant fixes for the same bug.

rgarg1:
@SMahil @debloper 
Even though the PR #2529 is approved based on UX suggestions, on second thought, I disapprove of UX suggestions. When it comes to Scenarios, Experiences etc, the entire section is clickable, but we break the user exp and restrict them to click on a limited area to select iterations. Why?

debloper:
> When it comes to Scenarios, Experiences etc, the entire section is clickable, but we break the user exp and restrict them to click on a limited area to select iterations. Why?

@rgarg1 group blocks don't have tree-expander or kebab menus, iteration blocks do. UX can't be same for both these section for this reason (unless we make drastic UX revamp). Which is why, unless we have a better proposal from anyone, this seems to be the best way around and is approved by UX now.

If you happen to have a better solution (there can always be a better solution), please do propose with visual representations.

rgarg1:
I can confirm that the clickable area is much larger now, even if not ideal. LGTM

PR #2570, opened 2018-04-19
fix(CI/CD): Fix failing chrome installation on CentOS

jarifibrahim:
Chrome installation on CentOS based containers is failing with
```
Package: google-chrome-stable-66.0.3359.117-1.x86_64 (google-chrome)
           Requires: libappindicator3.so.1()(64bit)
```

This PR adds `epel-release` and `libappindicator` packages which will fix the error.


jarifibrahim:
[test]

rgarg1:
@jarifibrahim I believe this only a stop-gap measure and not a permanent change.

jarifibrahim:
@rgarg1 This is a permanent solution. We're installing the `libappindicator` package which chrome needs.

PR #2593, opened 2018-04-26
fix(workitem): update workitem Title Bold value on all workitem edits

sahil143:

Update/Add a Workitem:
  workitem title will be bold if it's matching the current filter

### [ ] What issue/task does this PR references? _[Mandatory]_

https://github.com/openshiftio/openshift.io/issues/2823

### [ ] Are the tests Included? _[Mandatory]_

**YES**


src/app/effects/work-item.effects.ts
```diff
@@ -80,13 +83,14 @@ export class WorkItemEffects {
           wItem.createId = createID;
           return wItem;
         })
+        .switchMap(w => util.workitemMatchesFilter(this.route.snapshot, this.filterService, this.workItemService, w))
```
vikram-raj: Why are you passing the filterService and workItemService here? You can directly import it into util and use it.

src/app/effects/work-item-utils.ts
```diff
@@ -0,0 +1,53 @@
+import { Observable } from 'rxjs';
+import { WorkItemUI } from '../models/work-item';
+
+export function workitemMatchesFilter(route, filterService, workItemService, w): Observable<WorkItemUI> {
```
vikram-raj: What is ```w``` here? Give some meaningfull name.

sahil143: `w` is workitem here. I will change it to `workItem`

sahil143:
@Raunak1203 We can do this and we would need a proper notification message for this. Although, if a workitem is not matching filter we have been showing its title light. So, I think change in bold value is enough to indicate that workitem is not matching current filter anymore.
@nimishamukherjee

src/app/effects/work-item.effects.ts
sahil143: @vikram-raj To inject the services in utill i would need to create a class for this and I only needed services for this method only. So, passed the instances of these services through method call.

nimishamukherjee:
@sahil143 @Raunak1203 we can use the bold (to denote matching) and unbold (to denote not matching) items.

src/app/effects/work-item.effects.ts
vikram-raj: @sahil143 Ok got it.

nimishamukherjee:
[test]

PR #2598, opened 2018-04-30
fix(CI/CD) : fix failing chrome installation

Raunak1203:
Chrome installation on CentOS is failing beacuse chrome stable channel is down.
This PR adds chrome beta channel which will fix the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,7 @@ RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-
 
 COPY runtime/tests/google-chrome.repo /etc/yum.repos.d/google-chrome.repo
 
-RUN yum install -y google-chrome-stable
+RUN yum install -y google-chrome-beta
 
 ENV FABRIC8_USER_NAME=fabric8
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -52,7 +52,7 @@ RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-
 
 COPY runtime/tests/google-chrome.repo /etc/yum.repos.d/google-chrome.repo
 
-RUN yum install -y google-chrome-stable
+RUN yum install -y google-chrome-beta
```
rgarg1: So, we'll never run tests using stable chrome henceforth?

jarifibrahim: @rgarg1 I'm not sure what difference beta and stable have. We could revert to the stable version later on. Right now the installation channel for `chrome-stable` is broken. We cannot install latest stable version of chrome via the official channel.

rgarg1: @jarifibrahim Well Beta as the name suggests is beta quality! Do I need to say more?
Accepting this PR as a temporary fix.

PR #2551, opened 2018-04-12
feat(infotips): Added info tips in planner

divyanshiGupta:
Info tips User Story: https://openshift.io/openshiftio/openshiftio/plan/detail/1744
Test: WIP

Notes:
- Currently, info tips are hardcoded in the frontend as the backend is not ready, once the backend is ready hardcoded info tips will be removed.
- For few UI elements, info tips were not present in the JSON doc so I have used lorem ipsum text in its place

PR #2608, opened 2018-05-03
feat(infotips): Added Infotips in planner

divyanshiGupta:
Previous work which is merged: https://github.com/fabric8-ui/fabric8-planner/pull/2551.

Due to some architectural concerns raised in this issue https://github.com/fabric8-services/fabric8-wit/issues/2032 we are moving from backend support for infotip to frontend. 

The changes introduced by this PR - 
- Added service for infotip which calls the docs API - https://docs.openshift.io/json/infotips.json.
- Created store, actions, reducers and effects for infotip json data.
- Every infotip now gets the infotipText from the store based on its corresponding elements UUID.


src/app/components_ngrx/infotip/infotip.component.ts
```diff
@@ -1,12 +1,12 @@
-import { Component, Input, HostListener } from '@angular/core';
+import { Component, Input} from '@angular/core';
 
 @Component({
   selector: 'infotip',
-  template: `<template #popoverTemplate>
+  template: `<ng-template #popoverTemplate>
```
SMahil: make HTML template for infotips

SMahil: Also, infotips css is not here. Please add it here as template specific CSS rather than in assets folder

PR #2611, opened 2018-05-03
fix(work-item-detail): Use the state value sent from the backend and display it as is on the UI (git-3298)

nimishamukherjee:
**What does this PR do?** 
Use the state value sent from the backend and display it as is on the UI

**What issue/task does this PR references?**
https://github.com/openshiftio/openshift.io/issues/3298

**Are the tests Included?**
No


nimishamukherjee:
@Raunak1203 the UI will display the state sent by the backend.
I have raised an issue https://github.com/openshiftio/openshift.io/issues/3448 for the backend to maintain consistent cases for states.

Raunak1203:
ok @nimishamukherjee

Raunak1203:
Ohh Yes! Thanks for the clarification :+1:

PR #2646, opened 2018-05-16
Data normalisation initial work / demonstration 

sanbornsen:
In this PR we have normalized the user data. Now we have a state called `users` under `listPage` which holds the data **not** in a form of an array but a form of a normalized object. 
For example, if previously the data used to look like - 
```
const users = [{
      id: '1',
      name: 'Sudipta',
      add: 'Alpine Eco'
    }, {
      id: '2',
      name: 'Ibrahim',
      add: 'Brigade'
    }]
```
Now they look like 
```
const users = {
      '1': {
        id: '1',
        name: 'Sudipta',
        add: 'Alpine Eco'
      },
      '2': {
        id: '2',
        name: 'Ibrahim',
        add: 'Brigade'
      }
    }
```

**What does it solve?**
Now we don't copy the data to resolve it for some other attribute. For example consider `comment`, where `creator` is a relationship that relates to `user`. Previously, after fetching the `comment` we used to resolve the `creator` from the list of `users` and simply copy it to the creator section. Something like this - 
```
const comment = {
    id: 'comment-0',
    cretor: {
        id: '1',
        name: 'Sudipta',
        add: 'Alpine Eco'
      }, 
      message: 'This is a comment'
}
```
and there are various problem with this model, first, we are redundending the data, second comment will have no clue if the user's name is changed, third we rely only on the list of users that is fetched previously thus a non-collborator commentor's name and avatar is not shown right now (which is fixed in this implementation) and last but not the least in order to update the `user` the whole DOM tree of comment will be effected. 
So, now the comment data looks something like this - 
```
const comment = {
    id: 'comment-0',
    cretor: this.store.select('listPage').select('users').select(creator_id), // a store observable instance
    message: 'This is a comment'
}
```
Adn all we have to do in template to print the user name is 
``` 
<div *ngIf="comment.creator | async as creator">{{ creator.name }}</div>
 ```
Since, `comment.creator` is now an observable it's easy to watch over any change on that particular user and it'll update just this part of the DOM without effecting the entire comment section.

## Update

I recently had a question in mind, if storing the `Observable` into the store is a good idea? I checked with people from ngrx with this https://github.com/ngrx/platform/issues/1076 and turned out it's not. 

So we are not going to do the following - 
```
const comment = {
    id: 'comment-0',
    cretor: this.store.select('listPage').select('users').select(creator_id), // a store observable instance
    message: 'This is a comment'
}
```
Instead, we are going to keep the raw id of the relations like the following 

```
const comment = {
    id: 'comment-0',
    creator: creator_id
    message: 'This is a comment'
}
```
**But then are we going to resolve the data in the component?**
No, because these **queries are common for multiple components**, besides **we want to keep our components clean from `Store` import**. So I am adding **Injectable Query classes** to the attribute models.
For example, in this PR you'll find - 
`UserQuery` -> https://github.com/fabric8-ui/fabric8-planner/pull/2646/commits/1d809af356808b2f0e2ce0807f89a77caa6b8a8a#diff-b8f872575d850524306aa76adff1ca6eL76
`CommentQuery` -> https://github.com/fabric8-ui/fabric8-planner/pull/2646/commits/229ddd873b20071a781be36c70004157bf6e302f#diff-30f2bcacbe12a755d8c282d5a2a4498fR129
These query classes will be responsible for resolving and calculating query values for any attributes. 

## References - 
* https://netbasal.com/querying-a-normalized-state-with-rxjs-in-angular-71ecd7ca25b4
* https://netbasal.com/lets-talk-about-select-and-reselect-in-ngrx-store-177a2f6045a8

### DIFF
--- a/src/app/actions/user.actions.ts
+++ b/src/app/actions/user.actions.ts
@@ -1,15 +1,17 @@
+
 import { Action } from '@ngrx/store';
 import { UserUI } from './../models/user';
 import { UserState } from './../states/user.state';
 
 export const SET = '[users] Set';
 export const GET = '[users] Get';
+export const GET_ERROR = '[users] GetError';
 
 /**
  * This action class set the normalized user data
  * Be it one user or number of users we can use
  * this action to update the user state
- * The payload value is a disctionary of users
+ * The payload value is a dictionary of users
  * where the id is the key and value is the entire
  */
 export class Set implements Action {
@@ -32,5 +34,9 @@ export class Get implements Action {
   readonly type = GET;
 }
 
+export class GetError implements Action {
+  readonly type = GET_ERROR;
+}
+
 export type All
-  = Set | Get;
+  = Set | Get | GetError;
--- a/src/app/actions/user.actions.ts
+++ b/src/app/actions/user.actions.ts
@@ -1,0 +1,23 @@
+import { Action } from '@ngrx/store';
+import { UserUI } from './../models/user';
+import { UserState } from './../states/user.state';
+
+export const SET = '[users] Get';
+
+/**
+ * This action class set the normalized user data
+ * Be it one user or number of users we can use
+ * this action to update the user state
+ * The payload value is a disctionary of users
+ * where the id is the key and value is the entire
+ */
+export class Set implements Action {
+  payload: UserState;
+  constructor(payload: UserState) {
+    this.payload = payload;
+  }
+  readonly type = SET;
+}
+
+export type All
+  = Set
--- a/src/app/actions/user.actions.ts
+++ b/src/app/actions/user.actions.ts
@@ -2,7 +2,8 @@ import { Action } from '@ngrx/store';
 import { UserUI } from './../models/user';
 import { UserState } from './../states/user.state';
 
-export const SET = '[users] Get';
+export const SET = '[users] Set';
+export const GET = '[users] Get';
 
 /**
  * This action class set the normalized user data
@@ -19,5 +20,17 @@ export class Set implements Action {
   readonly type = SET;
 }
 
+/**
+ * This action is used to get one user
+ * from server by it's ID
+ */
+export class Get implements Action {
+  payload: string;
+  constructor(payload: string) {
+    this.payload = payload;
+  }
+  readonly type = GET;
+}
+
 export type All
-  = Set
+  = Set | Get;
--- a/src/app/effects/comment.effects.ts
+++ b/src/app/effects/comment.effects.ts
@@ -1,7 +1,6 @@
 import { Store } from '@ngrx/store';
 import { Actions, Effect } from '@ngrx/effects';
 import { Injectable } from '@angular/core';
-import { cloneDeep } from 'lodash';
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

**Comment by divyanshiGupta** [comment]
@sanbornsen comment model test is failing. Although the code looks good to me at the first look, will spend some more time on this tomorrow.

File: src/app/effects/comment.effects.ts [hunk 1]
```diff
@@ -1,15 +1,15 @@
 import { Store } from '@ngrx/store';
 import { Actions, Effect } from '@ngrx/effects';
 import { Injectable } from '@angular/core';
+import { cloneDeep } from 'lodash';
```
vikram-raj: Why you import ```cloneDeep```? As I see you are not using it anywhere in this file.

src/app/effects/user.effects.ts
```diff
@@ -0,0 +1,48 @@
+import { Actions, Effect } from '@ngrx/effects';
+import { Injectable } from '@angular/core';
+import {
+  Notification,
+  Notifications,
+  NotificationType
+} from 'ngx-base';
```
vikram-raj: You are not using notification anywhere in this file. So, we don't need this import.

michaelkleinhenz:
Possibly unrelated question: why does the store use ui-centric addressing: ´´´this.store.select('listPage')´´´? Is there a different user list for example on the board view?

sanbornsen:
> why does the store use ui-centric addressing: 

A very valid question. There is no reason and **I should move that out to root level**. The addressing are ui feature/page-centric considering there will be feature stores in children level, but we are not going too deep in it, therefore we can make them attribute oriented. That change is easy.

sanbornsen:
@michaelkleinhenz I later realized that we can't keep the `user` out of `listPage` because `listPage` is the root level feature in the planner. Ideally, this should go to the platform and be part of the state in AppModule. Instead, it's a part of a feature module now.

sanbornsen:
[test]

Raunak1203:
[test]

Raunak1203:
[test]

Raunak1203:
[test]

sanbornsen:
[test]

Raunak1203:
This Issue is in prod-preview as well, i've created a separate issue for this https://github.com/openshiftio/openshift.io/issues/3601

src/app/actions/user.actions.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const SET = '[users] Set';
+export const GET = '[users] Get';
+export const GET_ERROR = '[users] GetError';
+
+/**
+ * This action class set the normalized user data
+ * Be it one user or number of users we can use
+ * this action to update the user state
+ * The payload value is a disctionary of users
```
vikram-raj: Typo ~~~disctionary~~~ -> dictionary.

src/app/effects/user.effects.ts
vikram-raj: Now you are using it. So this comment not relevant anymore.

Issue #2506, opened 2018-03-20
Establish Makefile for building stuff on CICO and local machines


kwk:
As a follow-up to [this discussion](https://github.com/fabric8-ui/fabric8-planner/pull/2473#pullrequestreview-105307081):

@kwk:

> @pranavgore09 currently the file cico_run_tests.sh contains a bunch of install and other commands. We should separate those out and provide a shell-oriented Makefile that one can execute on a local machine as well. This way people test the same on local as on Jenkins CI. I do that all the time with fabric8-wit.

@debloper 

> We should separate those out and provide a shell-oriented Makefile that one can execute on a local machine as well.

@kwk

> agreed, but can that be considered into a followup patch? We (me and pranav) are considering addition of Makefile and define its most appropriate purpose (without taking-away/stepping-onto npm or gulp's tails). But that will need a bit of followup, and better done when this and the standalone Planner patch are through (so to say, the architecture is a bit more stable).

debloper:
Holding this back to allow architecture and build changes to go through/stabilize first (CICO and Standalone Planner). Working on this before those are through only increases the overall workload (back and forth changes).

PR #2651, opened 2018-05-21
feat(Makefile): use Makefile for running tasks, retiring gulp

debloper:
## What does this PR do?
This sets up the base to replace Gulp as a task runner, with Make. For the moment, both will work so that developers can get accustomed with the new set of commands. Once the transition phase is over, Gulp can be safely removed from repo and Make can take its place as user-run commands as well as in CI/CD routines.

## What issue/task does this PR references?
#2506 & https://openshift.io/openshiftio/openshiftio/plan/detail/2419

## Are the tests Included?
Not Applicable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ watch: watchjs watchcss watchhtml watchfiles
 
 clean: cleancache cleandist cleanimages cleanmodules cleantemp
 
+test: testunit
+
 processjs:
 	# Compiling TypeScript to JavaScript...
 	# (according to the configs set in default tsconfig.json)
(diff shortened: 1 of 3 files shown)


sanbornsen:
Can you write a readme for this? How do we build the planner with this setup?

debloper:
Took @jarifibrahim through the code and implementation, along with comparing it against the Gulpfile.
I believe he should be able to maintain this (if it's considered as a value addition) in my absence.

joshuawilson:
Since we are moving to a monorepo and this is really out of date, I'm closing it.

PR #2670, opened 2018-06-08
Fix peer dep build

joshuawilson:
This updates the changes to package.json so that build works.


joshuawilson:
wrong branch, closing

PR #2680, opened 2018-06-16
fix(dynamic-field): usability issues fixed with dynamic field

sanbornsen:

### Mandatory
 - [x] What does this PR do? 
     - onBlur the fields get closed and changes are lost
     - showing error message on input type mismatch
 
- [ ] Are the tests Included?
___
### Optional

### DIFF
--- a/src/app/widgets/inlineinput/inlineinput.component.html
+++ b/src/app/widgets/inlineinput/inlineinput.component.html
@@ -1,40 +1,59 @@
 <div class="inlineinput-container"
-[class.read-only]="disabled">
-<textarea
-  #input
-  autosize
-  (click)="startEditing($event, true)"
-  [init_height]="23"
-  [attr.disabled]="(!onLineClickEdit || disabled) && !editing ? '' : null"
-  [attr.placeholder]="placeholder"
-  [class.active]="editing"
-  [ngModel]="inputValue"
-  class="input-field inlineinput-textarea"
-  (keydown.enter)="submitOnEnter($event)"
-  (keydown)="onkeyDown($event)"
-  >
-</textarea>
-<div
-  *ngIf="editing"
-  (click)="saveClick()"
-  class="btn btn-primary action-btn btn-save inlineinput-btn-save"
-  [class.disabled]="saving">
-  <i class="fa fa-check"></i>
-</div>
-<div
-  class="btn btn-default action-btn inlineinput-btn-cancel"
-  *ngIf="editing"
-  (click)="closeClick()">
-  <i class="fa fa-close"></i>
-</div>
-
-<div class="editor-icon-container" *ngIf="!editing">
-  <i (click)="startEditing($event, false)"
-    class="pficon-edit edit-icon"></i>
-</div>
-<div *ngIf="editing">
-  <div class="error-message" *ngIf="errorMessage.length">
-    <small>{{ errorMessage }}</small>
+[class.read-only]="disabled"
+(clickOut)="closeClick()">
+  <textarea
+    #input
+    autosize
+    (click)="startEditing(true)"
+    [init_height]="23"
+    [attr.disabled]="(!allowOnLineClickEdit || disabled) && !editing ? '' : null"
+    [attr.placeholder]="placeholder"
+    [class.active]="editing"
+    [ngModel]="inputValue"
+    class="input-field inlineinput-textarea"
+    (keydown.enter)="submitOnEnter($event)"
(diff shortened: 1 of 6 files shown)


src/app/widgets/inlineinput/inlineinput.component.ts
```diff
@@ -139,6 +143,9 @@ export class InlineInputComponent implements OnInit {
           break;
       }
     }
+    if (this.isNotValid) {
+      this.errorMessage = `This is a ${this.type} field`;
+    }
   }
 
   checkInteger(keycode, event) {
```
jarifibrahim: This function won't work. You're checking for each `keycode` individually. `checkInteger` returns true for string like `1.3.4.5.`.
A better approach to validating input in the `onSave` method by using `parseInt`/`parseFloat` methods.

src/app/widgets/inlineinput/inlineinput.component.ts
```diff
@@ -77,11 +77,13 @@ export class InlineInputComponent implements OnInit {
   }
 
   closeClick() {
```
jarifibrahim: This method should validate the input value

sanbornsen: Why should it? I am closing the field here, not saving.

src/app/widgets/inlineinput/inlineinput.component.ts
sanbornsen: The motive was to show error when you type a character and don't see anything happening in the text field. Because we block keys to the respective fields.

src/app/widgets/inlineinput/inlineinput.component.ts
```diff
@@ -76,12 +86,24 @@ export class InlineInputComponent implements OnInit {
     }
   }
 
+  validateValue(value) {
+    if (this.type === 'integer') {
+      return /^\d+$/.test(value);
+    }
+    if (this.type === 'float') {
+      return /^-?\d*(\.\d+)?$/.test(value);
```
jarifibrahim: Why is negative value allowed flor float but not for integer?

src/app/widgets/inlineinput/inlineinput.component.ts
```diff
@@ -61,11 +67,15 @@ export class InlineInputComponent implements OnInit {
...
-      callBack: (v: string = '', e: string = '') => this.handleSave(v, e)
-    });
+    if (this.validateValue(this.inputField.nativeElement.value)) {
+      this.saving = true;
+      this.onSave.emit({
+        value: this.formatValue(this.inputField.nativeElement.value),
+        callBack: (v: string = '', e: string = '') => this.handleSave(v, e)
+      });
+    } else {
+      this.errorMessage = 'Invalid value for the field';
```
jarifibrahim: The error is very generic. We should make this error field specific

src/app/widgets/inlineinput/inlineinput.component.ts
jarifibrahim: A float can never have negative value because of this https://github.com/fabric8-ui/fabric8-planner/pull/2680/commits/43ab0dc0d9a4686db49bfe559adc924d728c080a#diff-ab69836b288ac3d161b883556bbbebc3R171 function. The input doesn't accept `-` character.

src/app/widgets/inlineinput/inlineinput.component.ts
jarifibrahim: >Because we block keys to the respective fields.

We shouldn't do this. Let the user type whatever he wants to and then throw the error while saving.

sanbornsen: That's a UX discussion.

sanbornsen: The point is the component is capable of doing the both.

src/app/widgets/inlineinput/inlineinput.component.html
```diff
@@ -1,40 +1,59 @@
...
-  <i (click)="startEditing($event, false)"
-    class="pficon-edit edit-icon"></i>
-</div>
-<div *ngIf="editing">
-  <div class="error-message" *ngIf="errorMessage.length">
-    <small>{{ errorMessage }}</small>
+  <!--
+    We can not put *ngIf in this component.
+    Understand the following flow -
+    1. You click on the edit icon
```
jarifibrahim: Thank you for adding the explanation

PR #2685, opened 2018-06-19
fix(planner-module): planner module exported

sanbornsen:

### Mandatory
 - [x] What does this PR do? 
    Right now we have three modules exported from the planner. The `PlannerListModule`, `PlannerBoardModule` and `WorkItemDetailModule`. Fabric8-ui is consuming them separately and putting on routers, which are configured in `app-routing.module` in `fabric8-ui`. The problem with this model is, sharing common data between these three modules. 

    In this update we are exporting one `PlannerModule` with the `/`, `/board`, `/detail/:id` router configuration in it. This way, `fabric8-ui` can cinsume one module from planner, put it in the right place and it should work as it is working right now. 
      - this module has list, board and details module
      - each of the exported modules has their own routing config
   
    #### The benefits we are getting here is:
      - `PlannerModule` can hold all the common data state
      - All the common components can be shared between list, board and detail module without re-instanciating for each of them.
      - Moving from one page to another inside planner will be faster since there is no resolver in between.

```
Note: This is backward compaitable. If fabric8-ui consume these changes without any change in the routing, nothing breaks.
```

    https://openshift.io/openshiftio/openshiftio/plan/detail/2593

- [ ] Are the tests Included?
___
### Optional


alien-ike:
### Ike Plugins (test-keeper)

Thank you @sanbornsen for this contribution!



For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin.

sanbornsen:
@sahil143 not part of this at this moment. When we have the board view ready you can just add it to the PlannerModule and it should start working on `/board`

PR #2698, opened 2018-06-27
fix(quickadd): Workitems can be added to the iterations now

divyanshiGupta:
Fixes https://github.com/openshiftio/openshift.io/issues/3859

Setting selected-iteration ID in the newly created work-item in this PR, that fixes the above issue i.e newly created work-item not getting added to the selected iteration.

Test [WIP]

### DIFF
--- a/src/app/components_ngrx/work-item-quick-add/work-item-quick-add.component.ts
+++ b/src/app/components_ngrx/work-item-quick-add/work-item-quick-add.component.ts
@@ -23,6 +23,7 @@ import { WorkItem, WorkItemRelations, WorkItemService } from '../../models/work-
 import { WorkItemTypeUI } from '../../models/work-item-type';
 import { IterationUI } from './../../models/iteration.model';
 import { WorkItemQuery } from './../../models/work-item';
+import { FilterService } from './../../services/filter.service';
 
 // ngrx stuff
 import { Store } from '@ngrx/store';
@@ -74,7 +75,8 @@ export class WorkItemQuickAddComponent implements OnInit, OnDestroy, AfterViewIn
     private route: ActivatedRoute,
     private renderer: Renderer2,
     private store: Store<AppState>,
-    private workItemQuery: WorkItemQuery) {}
+    private workItemQuery: WorkItemQuery,
+    private filterService: FilterService) {}
 
   ngOnInit(): void {
     this.createWorkItemObj();
@@ -147,6 +149,8 @@ export class WorkItemQuickAddComponent implements OnInit, OnDestroy, AfterViewIn
     if (event) {
       event.preventDefault();
     }
+    let currentRoute = cloneDeep(this.route.snapshot.queryParams);
+    let selectedIterationID = this.filterService.getConditionFromQuery(currentRoute.q, 'iteration');
     // Do we have a real title?
     // If yes, trim; if not, reassign it as a (blank) string.
     this.workItem.attributes['system.title'] =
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-ui/fabric8-planner/blob/1eab24ab37a241550cb0ac9ced8690704d5def65/.ike-prow/test-keeper.yaml).

src/app/components_ngrx/work-item-quick-add/work-item-quick-add.component.ts
```diff
@@ -74,7 +75,8 @@ export class WorkItemQuickAddComponent implements OnInit, OnDestroy, AfterViewIn
     private route: ActivatedRoute,
     private renderer: Renderer2,
     private store: Store<AppState>,
-    private workItemQuery: WorkItemQuery) {}
+    private workItemQuery: WorkItemQuery,
+    private filterService: FilterService) {}
```
sanbornsen: Can we use iterationQuery instead? You can get the selected iteration from there.

src/app/components_ngrx/work-item-quick-add/work-item-quick-add.component.ts
```diff
@@ -147,6 +149,8 @@ export class WorkItemQuickAddComponent implements OnInit, OnDestroy, AfterViewIn
     if (event) {
       event.preventDefault();
     }
+    let currentRoute = cloneDeep(this.route.snapshot.queryParams);
+    let selectedIterationID = this.filterService.getConditionFromQuery(currentRoute.q, 'iteration');
```
sanbornsen: If we have multiple iterations in the same query then this will likely fail.

divyanshiGupta: @sanbornsen  Can we have multiple iterations in the query? If yes, then which iterations wis will be shown in the list?

src/app/components_ngrx/work-item-quick-add/work-item-quick-add.component.ts
divyanshiGupta: I havent added workItemQuery and it is already being used `this.workItemQuery.getWorkItems()
        .filter(items => !!items.length)
        .subscribe(items => {
          // const addedItem = items.find(item => item.createId === this.createId);
          this.resetQuickAdd();
        })`

divyanshiGupta: I want the iteration id of the iteration the user clicks on and not the selected iteration.

divyanshiGupta: Selected iteration is one among the active iterations if a I use that all the newly created wis will be added to one selected iteration and not the iteration the user clicked on. I can change the variable name to something like `openIterationID` if  `selectedIterationID` is confusing.

src/app/components_ngrx/work-item-quick-add/work-item-quick-add.component.ts
divyanshiGupta: Since currently one wi can be associated to a single iteration then to which iteration the newly created wi will be added if we have multiple iterations in the query.

src/tests/specs/smoke/smokeTest.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { browser } from 'protractor';
 import { PlannerPage } from '../../page_objects/planner';
 import * as support from '../../support';
+import { WorkItem } from '../../ui/planner';
```
Raunak1203: @divyanshiGupta you can remove this import

src/tests/ui/planner/sidepanel.ts
```diff
@@ -1,6 +1,7 @@
 import { $, ElementFinder, by} from 'protractor';
 import  *  as ui from './../../ui';
 import * as support from '../../support';
+import { BaseElement } from './../../ui';
```
Raunak1203: you can remove this import and use ui.BaseElement here https://github.com/fabric8-ui/fabric8-planner/pull/2698/files#diff-7a240b018c48e08143bb41f696f3ec9bR92

divyanshiGupta:
@nimishamukherjee can we merge this?

PR #2701, opened 2018-07-02
fix(typography): fix typography units and align css w.r.t it

SMahil:
### Mandatory
 - [ ] What does this PR do? 
This PR fixes the typography units and avoid the consistent issue with overall OSIO.
With respect to changes, it will fix the broken CSS for the planner

https://github.com/openshiftio/openshift.io/issues/3825
- [ ] Are the tests Included?
NA
___

@rgarg1 @nimishamukherjee


rgarg1:
@SMahil 

Deployment page without visiting Planner:
fix_without_planner

Deployment page after visiting Planner:
fix_with_planner

The fonts in both cases appear the sam enow, so the issue is fixed.

PR #2706, opened 2018-07-03
fix(state): states are migrated from plannerList to PlannerModule

sanbornsen:

### Mandatory
 - [x] What does this PR do? 
    * This PR migrate all the states from PlannerListModule to PlannerModule
    * Changes are mostly in `PlannerModule` and `PlannerListModule`
    * All the `listPage` is replaced with `planner` in the state path.

- [x] Are the tests Included?
    No. This PR needs no test because it's Module level changes. No feature was added, method was created or modified.
    
___
### Optional

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -77,7 +77,7 @@ mach.tslint = function(src) {
   return gulp.src(src)
     .pipe(tslint({
       formatter: 'verbose',
-      configuration: 'tslint.json' 
+      configuration: 'tslint.json'
     }))
     .pipe(tslint.report())
 };
@@ -128,7 +128,10 @@ gulp.task('build', function (done) {
 
   // watch
   if (argv.watch) {
-    gulp.watch([appSrc + '/app/**/*.ts', '!' + appSrc + '/app/**/*.spec.ts']).on('change', function (e) {
+    gulp.watch([
+      appSrc + '/app/**/*.ts', '!' + appSrc + '/app/**/*.spec.ts',
+      appSrc + '/planner.module.ts'
+    ]).on('change', function (e) {
       util.log(util.colors.cyan(e) + ' has been changed. Compiling TypeScript.');
       mach.transpileTS();
     });
--- a/src/app/components_ngrx/planner-list/planner-list.module.ts
+++ b/src/app/components_ngrx/planner-list/planner-list.module.ts
@@ -39,14 +39,6 @@ import { WorkItemQuickAddModule } from './../work-item-quick-add/work-item-quick
 import { CookieService } from './../../services/cookie.service';
 import { FilterService } from './../../services/filter.service';
 
-// ngrx stuff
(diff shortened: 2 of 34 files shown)


bartoszmajsak:
How come we are approving such a large change without a single test? Especially that the template says it's mandatory.

sanbornsen:
@bartoszmajsak in this PR we changed the path for the test from `listPage` to `planner` thus it looks large. We don't write tests for modules because it only consists of imports and declaration.

bartoszmajsak:
@sanbornsen Then maybe justification on the mandatory step would help to clarify beforehand? Can this be done from now on, for every PR? Otherwise, I would suggest rethinking if we need such a template at all if we don't conform to it.

sanbornsen:
@bartoszmajsak justification on description makes sense. We need this template. A PR like this is very rare.

gulpfile.js
```diff
@@ -89,12 +89,12 @@ mach.transpileLESS = function (src, debug) {
     // paths: [ path.join(__dirname, 'less', 'includes') ]
   }
   return gulp.src(src)
-    .pipe(less({
-      plugins: [autoprefix]
-    }))
     .pipe(lesshint({
       configPath: './.lesshintrc' // Options
     }))
+    .pipe(less({
```
sahil143: Why the order was changed here for `.pipe(lesshint({` and `.pipe(less({`?

src/app/components_ngrx/planner-list/planner-list.module.ts
```diff
@@ -42,10 +42,6 @@ import { FilterService } from './../../services/filter.service';
 // ngrx stuff
 import { EffectsModule } from '@ngrx/effects';
```
sahil143: unused imports

src/app/components_ngrx/planner-list/planner-list.module.ts
```diff
@@ -42,10 +42,6 @@ import { FilterService } from './../../services/filter.service';
...
 import { StoreModule } from '@ngrx/store';
```
sahil143: same as above

sanbornsen:
@bartoszmajsak you can see, even we don't have enough unit tests overall, our functional tests are strong enough to identify problems. Somehow linking work item is broken in this PR and thus the build fails.

bartoszmajsak:
> @bartoszmajsak you can see, even we don't have enough unit tests overall, our functional tests are strong enough to identify problems. Somehow linking work item is broken in this PR and thus the build fails.

@sanbornsen Sure, except that it took ~30minutes to execute :) If you would have faster tests you potentially could spot this problem before opening the PR. Something to consider as an improvement moving forward.

sanbornsen:
@bartoszmajsak sure, couldn't agree more. But situation matters.

sudsen:
[test]

sanbornsen:
[test]

PR #2708, opened 2018-07-04
Board State and Actions

sahil143:

### Mandatory
 - [ ] What does this PR do? 
     This PR has actions states for Board View

- [ ] Are the tests Included?
Yes
____
### Optional

### DIFF
--- a/src/app/actions/board.actions.spec.ts
+++ b/src/app/actions/board.actions.spec.ts
@@ -1,0 +1,34 @@
+import { boardUIData } from '../services/board.snapshot';
+import { BoardState } from './../states/board.state';
+import {
+  Get, GET,
+  GET_ERROR, GET_SUCCESS,
+  GetError, GetSuccess
+} from './board.actions';
+
+describe('Unit Test :: Board Actions', () => {
+  it('GetBoard :: should create get action', () => {
+    const action = new Get('');
+    expect({...action}).toEqual({
+      type: GET,
+      payload: ''
+    });
+  });
+
+  it('GetBoardSuccess :: should create get success action', () => {
+    const boards = {
+      'board-1': boardUIData
+    } as BoardState;
+
+    const action = new GetSuccess(boards);
+    expect({...action}).toEqual({
+      type: GET_SUCCESS,
+      payload: boards
+    });
+  });
+
+  it('GetBoardError :: should create get error action', () => {
+    const action = new GetError();
+    expect({...action}).toEqual({type: GET_ERROR});
+  });
+});
--- a/src/app/actions/board.actions.ts
+++ b/src/app/actions/board.actions.ts
@@ -1,0 +1,34 @@
+import { Action } from '@ngrx/store';
+import { BoardState } from '../states/board.state';
+
+export const GET = '[board] Get';
+export const GET_SUCCESS = '[board] GetSuccess';
+export const GET_ERROR = '[board] GetError';
+
+export class Get implements Action {
+  payload: string;
+  constructor(payload: string) {
+    this.payload = payload;
+  }
+
+  readonly type = GET;
+}
+
+export class GetSuccess implements Action {
+  payload: BoardState;
+  constructor(payload: BoardState) {
+    this.payload = payload;
+  }
+
+  readonly type = GET_SUCCESS;
+}
+
+export class GetError implements Action {
+  readonly type = GET_ERROR;
+}
+
+export type All
+  = Get
+  | GetSuccess
+  | GetError;
+
--- a/src/app/actions/index.actions.ts
+++ b/src/app/actions/index.actions.ts
@@ -11,3 +11,4 @@ export { All as WIStateActions } from './work-item-state.actions';
 export { All as WITypeActions } from './work-item-type.actions';
 export { All as WorkItemActions } from './work-item.actions';
 export { All as EventActions } from './event.action';
+export { All as BoardActions } from './board.actions';
--- a/src/app/states/app.state.ts
+++ b/src/app/states/app.state.ts
@@ -5,6 +5,10 @@ export interface AppState {
 
   listPage?: ListPage;
 
+  boardView?: {
+    boards: states.BoardState
+  };
+
   iterationPanel: {
     iterationUI: states.IterationUIState
   };
@@ -25,7 +29,6 @@ export interface AppState {
     workItems: states.WorkItemState
   };
 }
-
 
 export interface ListPage {}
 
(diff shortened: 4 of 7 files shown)


src/app/actions/board.actions.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { boardUIData } from '../services/board.snapshot';
+import { BoardState } from './../states/board.state';
+import {
+  Get, GET,
+  GET_BOARD_URL, GET_BOARD_URL_ERROR,
+  GET_ERROR, GET_SUCCESS, GetBoardUrl, GetBoardUrlError, GetError, GetSuccess
+} from './board.actions';
```
vikram-raj: Can you please make it as ```import * as BoardActions from './board.actions'```.

src/app/effects/board.effect.ts
```diff
@@ -0,0 +1,71 @@
...
+    private boardService: BoardService,
+    private store: Store<AppState>,
+    private notifications: Notifications
+  ) {}
+
+  @Effect() getBoardApi: Observable<Action> = this.actions$
+    .ofType(BoardActions.GET_BOARD_URL)
+    .withLatestFrom(this.store.select(spaceSelector))
+    .switchMap(([actions, space]) => {
+      const url = space.relationships['space-template'].links.related;
```
vikram-raj: Change this variable name to ```spaceTemplateUrl```.

src/app/actions/board.actions.ts
```diff
@@ -0,0 +1,45 @@
+import { Action } from '@ngrx/store';
+import { BoardState } from '../states/board.state';
+
+export const GET = '[board] Get';
+export const GET_SUCCESS = '[board] GetSuccess';
+export const GET_ERROR = '[board] GetError';
+export const GET_BOARD_URL = '[board] GetBoardapiUrl';
```
sanbornsen: We don't need an action for this. The source of API URL is likely going to be changed from space template response to space response. We can make two consecutive service call from the effect to fetch board.

src/app/effects/board.effect.ts
```diff
@@ -0,0 +1,71 @@
...
+@Injectable()
+export class BoardEffects {
+  constructor(
+    private actions$: Actions,
+    private boardService: BoardService,
+    private store: Store<AppState>,
+    private notifications: Notifications
+  ) {}
+
+  @Effect() getBoardApi: Observable<Action> = this.actions$
```
sanbornsen: We don't need this extra chunk as I said earlier. `this.boardService.getBoardApiUrl(spaceTemplateUrl)` can be jammed into `getBoards` itself.

src/app/models/common.model.ts
```diff
@@ -144,13 +144,15 @@ export function cleanObject(obj: any, keysToRemove: string[] = []): any {
...
  */
-export function normalizeArray<I>(arr: I[]): {[id: string]: I} {
+export function normalizeArray<I>(arr: I[], id: string = ''): {[id: string]: I} {
   if (!Array.isArray(arr)) {
     throw(new Error('The input needs to be an array'));
   }
   let output = {};
   arr.forEach((item, index) => {
-    if (item.hasOwnProperty('id')) {
+    if (item.hasOwnProperty(id)) {
```
sanbornsen: Do you have a test in place for this in `common.model.spec.ts`? If no, then add it. Should be pretty simple.

sanbornsen:
## ToDo

- [ ] Add `workItemIds: string[]` to `BoardModelUI.columns[]` 
    - This will hold all the work item IDs in the column after query 
- [ ] Add another property to `workItemUI` model called `columnID` and `boardID`
    - You'll also need to write mapper for this.
- [ ] Add a reducer to `board.reducer` to update the work item IDs for each column 
    - This reducer listens to `'[workItem] GetSuccess'` event
    - Filters the item based on the `columnID`
    - stores ids in the respective column
- [ ] Add a new querry to work item query class to fetch items with a list of ids
    - It should look something like this 

```
getWorkItemsWithIds(ids: string[]): Store<WorkItemUI[]> {
  const selector = createSelector(
    workItemEntities,
    state => ids.map(i => this.resolveWorkItem(state[i]))
  );
  return this.store.select(selector);
}

And this can be used like this in your component

this.workItemQuery.getWorkItemsWithIds(column.workItemIds)
```
- [ ] create a query class in board model
   - Add `getBoard` query (follow work item query)

src/app/effects/board.effect.ts
```diff
@@ -0,0 +1,57 @@
...
+    .switchMap(([action, space]) => {
+      const spaceTemplateUrl = space.relationships['space-template'].links.related;
+      return this.boardService.getBoardApiUrl(spaceTemplateUrl);
+    })
+    .switchMap((url) => {
+      return this.boardService.getBoards(url)
+        .map(boards => {
+          const boardmapper = new BoardMapper();
+          return boards.data.map(board => {
+            board.relationships.columns.data.map(column => {
```
sanbornsen: This map data is not being used.

src/app/effects/board.effect.ts
```diff
@@ -0,0 +1,57 @@
...
+              return boards.included.filter(col => col.id === column.id);
```
sanbornsen: This is a O(n^3) operation could be done in O(n)

Normalize the columns part in `included` section and run map on that.

src/app/models/board.model.ts
```diff
@@ -73,21 +79,29 @@ export class BoardMapper implements Mapper<BoardModel, BoardModelUI> {
         fromPath: ['attributes', 'context'],
         toPath: ['context']
     }, {
-        fromPath: ['included'],
+        fromPath: ['relationships', 'columns', 'data'],
         toPath: ['columns'],
-        toFunction: (data) =>
-            Array.isArray(data) ? data.filter(d => d.type === 'boardcolumns') : []
+        toFunction: (data) => {
+            return data.map(col => {
```
sanbornsen: There is no check for the `data` being actually an array or not. If it sends null from the backend it'll fail.

src/app/services/board.service.spec.ts
```diff
@@ -61,7 +61,7 @@ describe('BoardService :: ', () => {
         boardService.getBoards('')
             .subscribe((url) => {
                 expect(url).toEqual(
```
sudsen: Why the response data is named as `url`

sahil143:
@sanbornsen I have made the changes you requested. As discussed, complexity will remain O(n^2) in the cases you mentioned.

PR #2714, opened 2018-07-05
fix(quick-preview/detail-page): Data being typed is lost on clicking outside.

divyanshiGupta:
### Mandatory
 - [x] What does this PR do? 
     Removed duplicate code for deactivating description editor.
     Moved `clickOut` directive to work-item-detail component from inline-input component to stop 
    `clickOut` from being applied on all the inline-inputs like dynamic fields and to be only applied on WI 
     title inline-input.

     Fixes https://github.com/openshiftio/openshift.io/issues/3842

- [x] Are the tests Included?
Yes
____
### Optional


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-ui/fabric8-planner/blob/94e6809c673835995dac39475a3a9672e1913606/.ike-prow/test-keeper.yaml).

rgarg1:
@divyanshiGupta The issue remains.

1) Make a change to a field
2) Move away from the field without saving it
3) Change Area or Iteration. Observe that the entered value in step 1 is lost.

divyanshiGupta:
@rgarg1 according to our discussion yesterday let us create a task for UX to decide upon if the dynamic fields should remain open on updating any other WI field.
I have also added the test please have a look.

src/tests/specs/agileTemplate.spec.ts
```diff
@@ -56,4 +56,16 @@ describe('Agile template tests: ', () => {
     await planner1.quickPreview.businessValue.untilTextIsPresentInValue('Business value for this Theme');
     expect(await planner1.quickPreview.businessValue.getAttribute('value')).toBe('Business value for this Theme');
   });
+
+  it('Dynamic fields should not get closed on outside click', async () => {
+    let newWorkItem = { title: 'Workitem of type Story', type : 'Storypoints'};
```
rgarg1: __Storypoints__ is not a valid Work item Type

src/tests/specs/agileTemplate.spec.ts
```diff
@@ -56,4 +56,16 @@ describe('Agile template tests: ', () => {
...
+    await planner1.createWorkItem(newWorkItem);
+    expect(planner1.workItemList.hasWorkItem(newWorkItem.title)).toBeTruthy();
+    /* Edit Story Points */
+    await planner1.workItemList.clickWorkItem(newWorkItem.title);
+    await planner1.quickPreview.storyPoints.clickWhenReady();
+    expect(await planner.quickPreview.isSaveButtonDisplayed()).toBeTruthy();
+    await planner1.quickPreview.labelDropdown.clickWhenReady();
+    expect(await planner.quickPreview.isSaveButtonDisplayed()).toBeTruthy();
```
rgarg1: Instead of the Save button, a better check would be to evaluate the value in the field.

rgarg1:
The PR is approved from the functionality perspective. The tests need to pass before the PR could be merged.

I will raise another UX task to track the case of dynamic fields clearing on WI reload.

Raunak1203:
@divyanshiGupta please do not remove the default template. Fill in the description and issue no. in the default template.

divyanshiGupta:
@Raunak1203 sure

Raunak1203:
Thanks @divyanshiGupta

src/app/widgets/inlineinput/inlineinput.component.html
```diff
@@ -1,6 +1,5 @@
 <div class="inlineinput-container"
-[class.read-only]="disabled"
-(clickOut)="closeClick()">
```
vikram-raj: Why you remove this from here and added it [here](https://github.com/fabric8-ui/fabric8-planner/pull/2714/files#diff-dbc24aee843bca8262d1c28fbaf591ccR142)?

divyanshiGupta: I removed outside click from here so that it doesn't apply on all the other inline-inputs like dynamic fields and so that it is only used for wi title inline-input

src/tests/ui/planner/workitem-quickpreview.ts
```diff
@@ -110,6 +110,11 @@ export class WorkItemQuickPreview extends ui.BaseElement {
   effortTextArea = new ui.TextInput(this.$('[placeholder="Effort"]'), 'effort textarea');
   workItemsGroup = new ui.Clickable(this.element(by.cssContainingText('alm-dynamic-field .f8-dynamic-control', ' Work Items ')),'Side panel WorkItem button');
   businessValue = new ui.TextInput(this.$('textarea[placeholder="Business Value"]'), ' Business value textarea');
+  storyPoints = new ui.TextInput(this.$('textarea[placeholder="Storypoints"]'), ' Storypoints textarea');
+  dynamicFieldDiv = new ui.BaseElement(this.$('.dynamic-integer-field'), 'Dynamic Integer Field Div');
```
Raunak1203: @divyanshiGupta W/element - more than one element found for locator By(css selector, .dynamic-integer-field) - the first result will be used
test is failing because of this please update dynamicFieldDiv value

PR #2751, opened 2018-08-09
fix(events): if oldValue or newValue not present return empty array

sahil143:

### Mandatory
 - [ ] What does this PR do? 
fixes: 
 - fix events according to backend changes
 - return empty array when oldValue or newValue is not present in relationships

https://github.com/openshiftio/openshift.io/issues/4147

- [ ] Are the tests Included?
___
### Optional


src/app/models/event.model.spec.ts
```diff
@@ -97,14 +95,99 @@ describe('Unit Test :: Events Model', () => {
...
+    expect(expectedOutPut).toEqual(output);
+  });
+
+  it('should correctly conver to ui model -3', () => {
+    const map = new EventMapper();
+    const input: EventService = {
+      'attributes': {
+        'name': 'system.assignees',
+        'newValue': null,
+        'oldValue': null,
```
kwk: These will no longer be present if the `name` refers to a relational field.

src/app/models/event.model.spec.ts
```diff
@@ -142,4 +225,74 @@ describe('Unit Test :: Events Model', () => {
     };
     expect(expectedOutPut).toEqual(output);
   });
+
+  it('should correctly conver to ui model - 5', () => {
```
kwk: typo: ~~conver~~ -> convert

kwk: same above and below

src/app/models/event.model.ts
```diff
@@ -178,7 +178,7 @@ export class EventQuery {
                   return this.areaQuery.getAreaObservableById(item.id);
                 })
               };
-            case 'system.assignees':
+            case 'users':
```
kwk: Cool!

src/app/models/event.model.spec.ts
sahil143: This is intentional. This test checks that the correct UI model is converted from a relational field with `null` values in attributes.

src/app/models/event.model.spec.ts
sahil143: fixed

sahil143:
> Looking good but you really should have a test that checks if you can ha.....

added a test for the same https://github.com/fabric8-ui/fabric8-planner/pull/2751/files#diff-54a28a37b7f272fea6972c13962b87eaR304

src/app/models/event.model.spec.ts
kwk: Sure but the keys (`newValue` and `oldValue`) will not be present (not event set to `null`) in the response if the change happened in a relational response. That's what I tried to say.

kwk: Btw. do you have tests for list values in `data.attributes.newValue` and `data.attributes.oldValue`? I couldn't find them anywhere. @vikram-raj ?

sahil143: >Sure but the keys (newValue and oldValue) will not be present (not event set to null) in the response if the change happened in a relational response. That's what I tried to say.

@kwk this is what you are looking for https://github.com/fabric8-ui/fabric8-planner/pull/2751/files/0e1f088d933b9691e111d941a404073c89b1134b#diff-54a28a37b7f272fea6972c13962b87eaR232.
I have added this test to ensure the current implementation(mapper function) doesn't break regardless attributes values are not present(or set to `null`) in response.

sahil143: > Btw. do you have tests for list values in data.attributes.newValue and data.attributes.oldValue? I couldn't find them anywhere. @vikram-raj ?

No, We dont have tests for list values `data.attributes.newValue` and `data.attributes.oldValue` because we dont have such field.

kwk: @sahil143 you should add those tests because those fields will come. We currently support them in the backend as well.

kwk: It is never a good thing to implement only against what we have but what we support.

sahil143: agreed!

sahil143: @kwk added tests and support for list values in `data.attributes` https://github.com/fabric8-ui/fabric8-planner/pull/2751/commits/d59bcc19b31e3e1fd2bb520c3b5ace3c574b2a2d

src/app/models/event.model.spec.ts
```diff
@@ -142,4 +225,184 @@ describe('Unit Test :: Events Model', () => {
...
+          type: 'users'
+        }
+      ],
+      oldValueRelationships: [
+        {
+          id: '20694424-0841-4d6c-bfb5-bbbb0391b8db',
+          type: 'users'
+        }
+      ],
+      type: null
```
kwk: It seems like the work item type from the payload (`00000000-0000-0000-0000-000000000003`) is not parsed here but you need this information.

sahil143: @kwk type is not parsed here but on a later stage, before events are stored in `ngrx/store`.

kwk: @sahil143 okay fine as long it is taken the work item type that is given by the event that is fine. When rendering events you must know at each point in time what the type of the work item was at that given time. In the near future @jarifibrahim will introduce type changes and then you might have to do a lookup to find out what field was actually changed.

PR #2752, opened 2018-08-10
fix(test): Change spec timeout from 60 mins to 3 mins

jarifibrahim:
60 mins is too much for each spec. 
I have reduced the timeout from 60 mins to 3 mins

### DIFF
--- a/tests/protractor.conf.ts
+++ b/tests/protractor.conf.ts
@@ -27,7 +27,7 @@ let conf: Config = {
     showColors: true,
     silent: true,
     isVerbose: true,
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @jarifibrahim for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-ui/fabric8-planner/blob/4547fbc41a618148b3f6c7667b321ffe76723a14/.ike-prow/test-keeper.yaml).

PR #2755, opened 2018-08-14
fix(infotip): fetch and display non-static infotips

divyanshiGupta:

### Mandatory
 - [x] What does this PR do? 
        This PR adds support for non-static infotips ( can be changed by the user ).

        https://github.com/openshiftio/openshift.io/issues/4270
- [x] Are the tests Included?
       Yes
___
### Optional

### DIFF
--- a/src/app/models/work-item-type.ts
+++ b/src/app/models/work-item-type.ts
@@ -38,9 +38,6 @@ export class WorkItemType extends modelService {
       guidedChildTypes?: {
         data?: WorkItemType[]
       },
-      infoTip?: {
-        data?: string;
-      },
       space?: Space
     };
 }
@@ -87,9 +84,6 @@ export class WorkItemTypeMapper implements Mapper<WorkItemTypeService, WorkItemT
         fromPath: ['attributes', 'version'],
         toPath: ['version']
       }, {
-        fromPath: ['attributes', 'description'],
-        toPath: ['description']
-      }, {
         fromPath: ['relationships', 'guidedChildTypes', 'data'],
         toPath: ['childTypes'],
         toFunction: (item: WorkItemTypeService) => {
@@ -106,8 +100,8 @@ export class WorkItemTypeMapper implements Mapper<WorkItemTypeService, WorkItemT
         toPath: ['type'],
         toValue: 'workitemtypes'
       }, {
-        fromPath: ['relationships', 'infotip', 'data'],
-        toPath: ['infotip'],
+        fromPath: ['attributes', 'description'],
+        toPath: ['description'],
         toFunction: function(value) {
           if (value === null) {
             return 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.';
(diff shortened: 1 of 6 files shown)


src/app/models/work-item-type.ts
```diff
@@ -89,8 +83,8 @@ export class WorkItemTypeMapper implements Mapper<WorkItemTypeService, WorkItemT
         toPath: ['type'],
         toValue: 'workitemtypes'
       }, {
-        fromPath: ['relationships', 'infotip', 'data'],
-        toPath: ['infotip'],
+        fromPath: ['attributes', 'description'],
+        toPath: ['description'],
```
sanbornsen: As we are sure of description coming from from the backend, do we need to give a default text anymore? We can still check for `null` and return empty string on `null`.

divyanshiGupta: Yes we can do that.

divyanshiGupta:
@rgarg1 I have added the task reference

src/app/models/work-item-type.ts
```diff
@@ -106,11 +100,11 @@ export class WorkItemTypeMapper implements Mapper<WorkItemTypeService, WorkItemT
...
         toFunction: function(value) {
```
sanbornsen: As we are not doing much here, let's compact the code, `(value) => value ? value : ''`  This will save 6 lines.

rgarg1:
@divyanshiGupta The infotip for the _Work item_ in the left panel is empty and is displayed a blank. Why so?

src/app/models/work-item-type.ts
rgarg1: Why not a placeholder text like "coming soon..." instead of empty info-tip that may look like an error. Let's see what UX has to suggest. @Veethika

divyanshiGupta:
@rgarg1 we have two methods for fetching infotips in planner as there are two types of infotip, static like iteration and non-static like work-item-types hence in the first method we fetch static infotips from the infotips doc and to let the docs team know which infotip is missing we show the UUID and in the second method we fetch non-static infotips from backend and since the infotip for work-items is null, empty string is being displayed once work-item infotip is added in backend it will reflect in the UI, @Preeticp did you add infotip for the work-items group type? 

@rgarg1 last time I talked to @Preeticp about this so she told me that the prod for infotips doc was not updated as the build failed. That is why it is showing the UUID for iteration, once prod will get updated the infotip will be shown. @Preeticp is prod still not updated?

rgarg1:
@divyanshiGupta 

The UUID for the info tip for _work item_ is showing up in prod-preview. Let's hear from @Preeticp before closing this.

Preeticp:
@rgarg1 @divyanshiGupta docs are still not getting updated in Prod, as per @rkratky in our last meeting. He has raised a concern about this, hopefully it will be resolved soon.

rgarg1:
LGTM

divyanshiGupta:
@Preeticp work-items group-type in agile template still doesnt have a infotip. Please add this too in the yaml.

Preeticp:
@divyanshiGupta wouldn't the infotip for this be taken from the json? The glossary and the json have been updated with the correct UUID for work item: https://github.com/fabric8io/fabric8-online-docs/blob/master/json/infotips.json

divyanshiGupta:
@Preeticp it wouldn't be taken from the json doc as this is a group-type and for group types we fetch infotips from backend and that is why we need work-item infotip in backend, please add it there.

rgarg1:
No more empty info-tip now. Fixed.

PR #2758, opened 2018-08-16
fix(query): fix provider dependencies for local env using feature-flag

divyanshiGupta:

### Mandatory
 - [ ] What does this PR do? 
- Due to some npm dependency issue in `npm link` feature-flag was not working locally when linking fabric8-planner with fabric8-ui, but it was working fine for prod builds on centOS as well as local fabric8-ui with the latest planner@0.53.2. 
- So in this PR fixes this issue by providing `FeatureFlagResolver`and other FeatureFlag related dependencies explicitly. 
- It also provides `FABRIC8_FEATURE_TOGGLES_API_URL` explicitly by using `process.env.FABRIC8_WIT_API_URL` as a factory since toggle service uses `WIT_API_URL` as a proxy. 
- This patch works both locally and with prod builds.

- [ ] Are the tests Included?
___
### Optional

### DIFF
--- a/src/app/components_ngrx/planner-query/planner-query.module.ts
+++ b/src/app/components_ngrx/planner-query/planner-query.module.ts
@@ -3,6 +3,9 @@ import { NgModule } from '@angular/core';
 import { PlannerQueryRoutingModule } from './planner-query-routing.module';
 import { PlannerQueryComponent } from './planner-query.component';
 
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-ui/fabric8-planner/blob/060c24d21e5797125dd16fc9b2e48947f2f3e5b4/.ike-prow/test-keeper.yaml).

PR #2775, opened 2018-08-30
fix(CI/CD): Fix failing master build

jarifibrahim:
The planner master build (https://ci.centos.org/view/Devtools/job/devtools-fabric8-planner-npm-publish-build-master/) has been failing for a while. The planner package is released but the build fails at creating the Pull Request to fabric8-npm-dependencies repository. 

With this PR 
 - The master build ends at semantic release step. **NO** pull request will be raised to `fabric8-npm-dependencies` or `fabric8-ui`.


cico_release.sh
```diff
@@ -16,9 +16,11 @@ function main() {
     # Build and Release Planner (It will update the tag on github and push fabric8-planner to npmjs.org)
     npm run semantic-release
 
-    create_merge_PR
 }
 
+#
+#   This function no longer works.
+#
```
nimishamukherjee: As we are not calling the function any more, can we remove it?

jarifibrahim: We could but I thought it might be useful to still keep it.
But I'll remove it. The function that doesn't work is useless :+1:

sanbornsen: Well, the difference is we don't just commit the `package.json` also we create a new `package-lock.josn` and then raise the PR. If that process could be automated then it'd be great. I'd rather keep the function and leave a comment to do that part inside in future.

PR #2798, opened 2018-11-01
fix(dynamic-component): Remove None from Enum field values

jarifibrahim:
Fixes https://github.com/openshiftio/openshift.io/issues/3831 and https://github.com/openshiftio/openshift.io/issues/3832#issuecomment-433874194

This PR removes the unnecessary `None` field in the dropdown for enum types. The backend would always send a default value for enum fields.

**Before**
>

**Now**
>    screenshot from 2018-11-01 17-26-20

### DIFF
--- a/src/app/components_ngrx/dynamic-field/dynamic-field.component.html
+++ b/src/app/components_ngrx/dynamic-field/dynamic-field.component.html
@@ -22,7 +22,6 @@
       <!-- workitemtype field type: dropdown -->
       <common-selector
         *ngIf="fieldValue.field.type.kind==='enum'"
-        [noValueLabel]="'None'"
         [allowUpdate]="editAllow"
         [headerText]="fieldValue.field.label + ' dropdown'"
         [items]="dropdownMenuItems"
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @jarifibrahim for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-ui/fabric8-planner/blob/90aef3b6a84b0c8076dc548c5e71329a0873bdd1/.ike-prow/test-keeper.yaml).

jarifibrahim:
[test]

kwk:
> /ok-without-tests

@sahil143 how can this be okay without a test?

jarifibrahim:
>@sahil143 how can this be okay without a test?

@kwk We don't have the tests for the dropdowns. We test user interactions via `e2e tests` and we test `services` via `unit test`. The piece of code I modified doesn't have any test (and it's not not easy to test it because it changes some HTML. You can't easily test rendered HTML)

jarifibrahim:
> ind a way to test it though. Relying on tests that run in a different repo is not very cool because you cannot verify this. If everything was in monorepo things looks different I guess.

@kwk No no. We have our own `e2e tests`. So the PR build executed the `unit tests` and about 68 tests `functional Tests (aka e2e tests)` on each SDD and AGILE template. We cover most of the user interaction scenario via these e2e-tests.

The QE team has their own e2e tests which cover the entire osio workflow and we have our own e2e tests that cover common planner workflows.

PR #2803, opened 2018-11-12
fix(board): Rename system.boardcolumns to boardcolumns

jarifibrahim:
Once https://github.com/fabric8-services/fabric8-wit/pull/2345 is merged the `system.boardcolumn` the relationship will be renamed to `boardcolumns`.

This PR makes necessary changes on the UI to address the changes in the backend.

### DIFF
--- a/src/app/models/work-item.spec.ts
+++ b/src/app/models/work-item.spec.ts
@@ -239,7 +239,7 @@ describe('WorkItemMapper', () => {
           'labels': {
             'data': []
           },
-          'boardColumns': {},
+          'boardcolumns': {},
           'workItemLinks': {
             'links': {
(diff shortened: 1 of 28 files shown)


src/app/models/work-item.spec.ts
```diff
@@ -239,7 +239,7 @@ describe('WorkItemMapper', () => {
           'labels': {
             'data': []
           },
-          'system.boardColumns': {},
+          'boardColumns': {},
```
kwk: @jarifibrahim why do you call this `boardColumns` and not `boardcolumns`?

jarifibrahim: I thought it's in camel case. I've changed it in https://github.com/fabric8-ui/fabric8-planner/pull/2803/commits/3399dfeee30353446153111446696adce91d9d86

joshuawilson: @kwk why not use camel case?

PR #2805, opened 2018-11-19
[WIP] Upgrade

joshuawilson:
This is updating all the libs.


alien-ike:
### Ike Plugins (test-keeper)

Thank you @joshuawilson for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-ui/fabric8-planner/blob/3877e76403df7294a3dcae065e49746fbb5a23a5/.ike-prow/test-keeper.yaml).

src/app/reducers/area.reducer.ts
```diff
@@ -5,7 +5,7 @@ import { AreaState, initialState } from './../states/area.state';
 
 export type Action = AreaActions.All;
 
-export const AreaReducer: ActionReducer<AreaState> = (state = initialState, action: Action) => {
+export function AreaReducer(state = initialState, action: Action) {
```
sanbornsen: I prefer exporting constants as arrow functions so that you can use the 'this' reference if needed anywhere. Well, reducers being pure functions it totally makes sense not to use any external reference hence this change is okay.
Can you add the return types to the functions though? i.e. for this one, it'd be - 
```
export function AreaReducer(state = initialState, action: Action):  ActionReducer<AreaState> {
```

joshuawilson: I do too but typescript / angular can't figure out what is going on here. The latest version doesn't transpile it correctly. If you want, reset the last 2 commits and you can see the error.

christianvogt: `this` will never be used here, so the change is fine however I an error because of the use of a big arrow function is very concerning considering it's a standard language construct.

joshuawilson: If you can find another way, I would be happy to take it. This is the only thing that allowed me to move forward. Just reset the last 2 commits and try to get it to build.

joshuawilson:
Since I couldn't get the build to compile and we are going to move this to the monorepo, I'm going to close this.

